PR #2, opened 2020-11-05
Update to SDK v0.12.1

turbolent:
Use the latest released version


psiemens (approved):

turbolent:
@psiemens @janezpodhostnik This is odd: The HelloWorld contract upgrade fails when the AST cache is enabled. Are we correctly handling cache invalidation in the FVM? I disable the cache for now

psiemens:
@turbolent 🤔 

Is this a regression, or is it failing due to previously untested functionality (i.e. multiple contracts per account)? 

AFAIK this was working correctly before. After searching through FVM and Cadence, I don't see any code that _explicitly_ handles cache invalidation after contract upgrades, but I did find this line, which caches a program in `parseAndCheckProgram` after checking and parsing: https://github.com/onflow/cadence/blob/3d3e3ed93f2068f0c5e7d55220e1a4a121508233/runtime/runtime.go#L566

The `parseAndCheckProgram` function is called during contract deployment, and it looks like that logic has changed in the runtime. Any chance that's the cause for this regression?

turbolent:
I looked into it and it turns out I introduced a regression while implementing multiple contracts per account: We previously had the issue where the cache was used when checking the program update, though it should not: https://github.com/onflow/cadence/commit/63044233f9fb799d1218d4f103bac9db80b1bf96. 

When I implemented multiple contracts per account I used the exported function again which uses the cache: https://github.com/onflow/cadence/blob/d114710e6759326b0d7bbeaf41f6c70ddefdef62/runtime/runtime.go#L1386

I'll push a fix and will release v0.10.1 of Cadence once merged.

go.mod
```diff
@@ -10,10 +10,10 @@ require (
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
 	github.com/improbable-eng/grpc-web v0.12.0
 	github.com/logrusorgru/aurora v0.0.0-20200102142835-e9ef32dff381
-	github.com/onflow/cadence v0.10.0
+	github.com/onflow/cadence v0.10.1
 	// this references: https://github.com/onflow/flow-go/tree/feature/multiple-contract-support.
-	github.com/onflow/flow-go v0.10.2-0.20201026185922-70cd25209148
-	github.com/onflow/flow-go-sdk v0.12.0
+	github.com/onflow/flow-go v0.11.1-0.20201109210205-312549711f62
```
psiemens, line 15: I think we should tag `onflow/flow-go` with a semantic version to avoid issues with missing commits in the future

turbolent: There is, `feature/multiple-contract-support`, as the comment mentions. Unfortunately, Go doesn't support pinning against branches or custom tag names. Once that feature PR is merged, we will switch back to flow-go's last released version.

psiemens: Oh I see, my bad. Makes sense!

psiemens (approved):

psiemens (approved):
Good idea to use `flowgo.DefaultMaxGasLimit`

turbolent:
@psiemens there's still a test failing: 

```
=== RUN   TestSubmitTransaction_EnvelopeSignature/Missing_envelope_signature
    TestSubmitTransaction_EnvelopeSignature/Missing_envelope_signature: transaction_test.go:431: 
        	Error Trace:	transaction_test.go:431
        	Error:      	Received unexpected error:
        	            	invalid address: 0000000000000002
        	Test:       	TestSubmitTransaction_EnvelopeSignature/Missing_envelope_signature
```

In `configureTransactionValidator` I'm already using `conf.GetChainID().Chain()`

turbolent:
@psiemens Fixed the address validation by enabling simple addresses. Could you please have another final look?

Along the way I saw some tests were TODOs and skiped – I tried to enable them (see 315b530) ... and they pass now :-)

psiemens (approved):
LGTM

PR #40, opened 2021-03-29
Update to Cadence v0.14.4, SDK v0.17.0, and Flow Go v0.15.3

turbolent:


headers.go
```diff
@@ -31,6 +31,18 @@ type headers struct {
...
+	panic("TODO")
+}
+
+func (h headers) BatchIndexByChunkID(headerID, chunkID flowgo.Identifier, batch storage.BatchStorage) error {
+	panic("TODO")
+}
+
+func (h headers) IDByChunkID(chunkID flowgo.Identifier) (flowgo.Identifier, error) {
+	panic("TODO")
+}
```
turbolent, line 44: @ramtinms @m4ksio Do these function need to be implemented for FVM to work? OK to leave them as no-ops, or is a simple in-memory storage of these needed?

ramtinms: I'm not sure if we need them, if you remove it and nothing complains I guess we don't need them

turbolent: They're required by the interface, I was just wondering if they also have to be properly implemented. Tests pass right now, but I was just wondering when/if this will be used, or maybe it isn't outside of the node software.

psiemens (approved):

PR #54, opened 2021-05-25
Return error if script fails

sideninja:
When a script execution fails error should be returned to the client, this is consistent with mainnet and also allows the developer to see what the error was in the client.

Error currently returned was an error of the cadence parser and did not provide any value or information to the client.


turbolent:
@sideninja Good catch! Could you please add a test for this? I think there's already a positive one

turbolent (approved):
Looks good! Thank you for adding the test 👍

PR #56, opened 2021-06-08
Bump `flow-go` version

jordanschalm:
## Description

Updates the pinned `flow-go` version to latest `master` (`v0.17.6`) in order to address recent breaking changes in `fvm`. These breakages are currently a blocker for the epochs feature branch on `flow-go`.

⚠️ I need some help with the configuration change required here. I'm not sure how to convert between the context config and the bootstrap config for restricted account deployment.

______

For contributor use:

- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work
- [ ] Re-reviewed `Files changed` in the Github PR explorer


blockchain.go
```diff
@@ -441,6 +440,7 @@ func configureBootstrapProcedure(conf config, flowAccountKey flowgo.AccountPubli
 	options := make([]fvm.BootstrapProcedureOption, 0)
 	options = append(options,
 		fvm.WithInitialTokenSupply(supply),
+		fvm.WithRestrictedAccountCreationEnabled(false),
```
jordanschalm, line 443: Tests fail with `missing account created event` with only this change.

turbolent:
> I need some help with the configuration change required here. I'm not sure how to convert between the context config and the bootstrap config for restricted account deployment.

@jordanschalm Can you elaborate on the problem? As in, what is the issue when using the Emulator without the configuration changes?

blockchain.go
janezpodhostnik: I can take a look. I know what the problem is!

janezpodhostnik:
The problem is that account creation emits more than one event now. The emulator tests assumed that there is only one event. This same assumption might be problematic in the go-sdk examples (and maybe elsewhere) this should be communicated as a breaking change. I will pass this on.

turbolent:
Oh, I think I understand now, thanks for the explanation 👍

turbolent (approved):
Can we maybe even bump to today's Canary spork version which has Cadence v0.17.0?

jordanschalm:
Hey @turbolent I updated again to the latest version. There was one small change from the Cadence version bump - could you take another look before I merge?

turbolent (approved):
Looks good, thank you for updating this! 👍

Issue #64, opened 2021-07-21
Flags for storage parameters


janezpodhostnik:
### Some context

Storage parameters are:
- amount of storage an account gets per FLOW
- and the minimum account balance (a.k.a.: minimum storage reservation)

They are defined on the `FlowStorageFees` contract here https://github.com/onflow/flow-core-contracts/blob/9a02fe98f39ef8c8632be88cc32c0ac6862fee50/contracts/FlowStorageFees.cdc#L28

If they get updated, they get updated with a transaction, not a code change.

The emulator currently takes the value of storage parameters from constants defined in `flow-go`: https://github.com/onflow/flow-go/blob/20969da4737c850ddc3cd43afdb741d1bcedc2f8/fvm/bootstrap.go#L55

### Problem definition

When storage parameters are changing, the emulator might not yet have the most recent version of `flow-go` (so storage constants are not updated yet), but users should be able to test how the change in storage parameters affects their code as soon as possible.

### Solution

Introduce optional flags (or environment variables) that can be set to change the storage parameters.

PR #65, opened 2021-07-21
Expose storage parameters

janezpodhostnik:
Closes: https://github.com/onflow/flow-emulator/issues/64

## Description

Introduced two new flags (or environment variables):

- `min-account-balance` flag controls  the minimum account balance (and consequentially account creation fees). example: `go run ./cmd/emulator/main.go --min-account-balance '0.04'`
- `storage-per-flow` flag sets the MB amount of storage capacity an account has per 1 FLOW token it has. example: `go run ./cmd/emulator/main.go --storage-per-flow '100.0'`

Both flags are strings that will be parsed as a cadence UFix64.

The defaults of both flags are taken from current version of `flow-go`. At the time of writing this task those values are:
- `min-account-balance  = 0.001`
- `storage-per-flow = 10.0`

## Side effect

I have reduced the default total token supply by a factor of 10. I did this otherwise the service account storage overflows if `storage-per-flow` is set to `100.00` which is the current mainnet value. This can be reverted after issue https://github.com/onflow/flow-core-contracts/issues/150 is resolved.
______

For contributor use:

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,8 @@ This command has several useful flags:
 - `--persist` Enable persistent storage (uses [Badger](https://github.com/dgraph-io/badger) key-value DB)
 - `--dbpath` Path to store database (default: `"./flowdb"`)
 - `--storage-limit` Enable limiting account storage use to their storage capacity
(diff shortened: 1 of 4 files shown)


turbolent (approved):
Nice!

sideninja (approved):
LGTM

PR #79, opened 2021-09-14
Adding debugging feature for failed signatures

### LINKED ISSUE #70: Adding debugging feature for failed signatures
### Issue To Be Solved

The emulator throws an error when it receives in an invalid transaction signature, but does not specify why the signature was invalid. Given that the emulator is a development tool, it should be able to perform an extra check to provide more diagnostics information to help the developer.

A transaction signature can be invalid for one of the following reasons:
1. signature was generated with the wrong private key
2. signature was generated by hashing the transaction with the wrong hash algorithm (e.g. on-chain account says SHA2, client used SHA3)
3. transaction was mutated after the signature was generated (e.g. sign the transaction, then set the gas limit, but submit the old signature)

Scenario 1 is difficult to report without knowing the original private key. 

Scenario 2 can be detected by attempting to verify the signature with each other hash algorithm.

Scenario 3 cannot be reliably detected, but the emulator _can_ print out a log of the transaction structure that was signed, which the developer can then compare against the transaction that their client prepares.

sideninja:
Closes #70 

## Description
The emulator throws an error when it receives in an invalid transaction signature, but does not specify why the signature was invalid. Given that the emulator is a development tool, it should be able to perform an extra check to provide more diagnostics information to help the developer.

Functionality:
- Detects and report wrong hashing for a selected account
- Report transaction structure for signature errors during the verbose option

______

For contributor use:

### DIFF
--- a/blockchain.go
+++ b/blockchain.go
@@ -26,6 +26,7 @@ import (
 	"github.com/onflow/flow-go/crypto/hash"
 	"github.com/onflow/flow-go/engine/execution/state/delta"
 	"github.com/onflow/flow-go/fvm"
+	fvmcrypto "github.com/onflow/flow-go/fvm/crypto"
 	fvmerrors "github.com/onflow/flow-go/fvm/errors"
 	"github.com/onflow/flow-go/fvm/programs"
 	"github.com/onflow/flow-go/fvm/state"
@@ -851,7 +852,6 @@ func (b *Blockchain) executeNextTransaction(ctx fvm.Context) (*types.Transaction
 			if err != nil {
 				return nil, err
 			}
-
 			return tx, nil
 		},
 	)
@@ -864,6 +864,12 @@ func (b *Blockchain) executeNextTransaction(ctx fvm.Context) (*types.Transaction
 	if err != nil {
 		// fail fast if fatal error occurs
 		return nil, err
+	}
+
+	// if transaction error exist try to further debug what was the problem
+	if tr.Error != nil {
+		tr.Error = b.debugErrorHashingAlgo(tr.Error, tp.Transaction)
+
 	}
 
 	return tr, nil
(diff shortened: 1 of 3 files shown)


server/backend/backend.go
```diff
@@ -608,6 +608,16 @@ func printTransactionResult(logger *logrus.Logger, result *types.TransactionResu
...
 		)
+
+		if sigErr, ok := result.Error.(*types.TransactionSignatureError); ok {
+			logger.
+				WithField("payer", sigErr.Transaction().Payer).
+				WithField("proposer", sigErr.Transaction().ProposalKey.Address).
+				WithField("proposerKeyIndex", sigErr.Transaction().ProposalKey.KeyIndex).
+				WithField("authorizers", sigErr.Transaction().Authorizers).
+				WithField("gasLimit", sigErr.Transaction().GasLimit).
+				Debug("❗  Transaction data")
```
janezpodhostnik, line 619: should this say `Transaction Signature Error`?

blockchain.go
```diff
@@ -1112,11 +1119,26 @@ func convertToSealedResults(
...
-// debugErrorHashingAlgo tries to unwrap error to the root and test for invalid hashing algorithms
-func (b *Blockchain) debugErrorHashingAlgo(err error, tx *flowgo.TransactionBody) error {
+// testGeneralSignatureError checks if there was signature error and add context for debugging
+func (b *Blockchain) testGeneralSignatureError(err error, tx *flowgo.TransactionBody) error {
+	flowErr, ok := err.(*types.FlowError)
+	if !ok {
+		return nil
+	}
+
+	_, ok = flowErr.Unwrap().(*fvmerrors.InvalidProposalSignatureError)
```
janezpodhostnik, line 1129: Could `errors.Is` or `errors.As` be used here? https://pkg.go.dev/errors

janezpodhostnik (commented):
Nice. I left some small questions.

server/backend/backend.go
sideninja: Agree

blockchain.go
sideninja: good idea, thanks

accounts_test.go
```diff
@@ -670,7 +672,10 @@ func TestRemoveAccountKey(t *testing.T) {
 	result, err = b.ExecuteNextTransaction()
 	assert.NoError(t, err)
 
-	unittest.AssertFVMErrorType(t, &fvmerrors.InvalidProposalSignatureError{}, result.Error)
+	unittest.AssertFVMErrorType(t,
+		&fvmerrors.InvalidProposalSignatureError{},
+		result.Error.(*types.SignatureError).Unwrap(),
+	)
```
turbolent, line 678: Instead of force-casting, use e.g. `require.ErrorAs` in a separate line before. Same in the the assertions below

sideninja: Good idea, thank you

janezpodhostnik (approved):

accounts_test.go
```diff
@@ -670,7 +671,8 @@ func TestRemoveAccountKey(t *testing.T) {
 	result, err = b.ExecuteNextTransaction()
 	assert.NoError(t, err)
 
-	unittest.AssertFVMErrorType(t, &fvmerrors.InvalidProposalSignatureError{}, result.Error)
+	var sigErr *fvmerrors.InvalidProposalSignatureError
+	assert.True(t, errors.As(result.Error, &sigErr))
```
turbolent, line 675: Maybe use the `assert.ErrorAs` assertion here, as it provides a better description when the assertion fails. Same in the other tests below

turbolent (approved):

PR #87, opened 2021-09-26
Emulator Feature: Report gas used

avcdsld:
Closes: FLIP Fest issue 12 https://github.com/onflow/flip-fest/issues/12

## Description

There should be a way to see transaction gas (computation) usage.

With this change, the output result of the transaction execution will look like this

**BEFORE**
スクリーンショット 2021-09-27 0 02 12

**AFTER**
スクリーンショット 2021-09-29 9 06 58

## Other

Since there was no specific test code for the standard output of "txID=xxxxxx" that is currently displayed, I decided that there was no need to add any specific test code for this change as well.

I checked to see if any changes were needed in this document, but it looks like no changes are needed.
https://docs.onflow.org/flow-cli/start-emulator/

______

For contributor use:

### DIFF
--- a/types/result_test.go
+++ b/types/result_test.go
@@ -35,21 +35,21 @@ func TestResult(t *testing.T) {
 		idGenerator := test.IdentifierGenerator()
 
 		trSucceed := &types.TransactionResult{
-			TransactionID:      idGenerator.New(),
-			ComputationGasUsed: 20,
-			Error:              nil,
-			Logs:               []string{},
-			Events:             []flow.Event{},
+			TransactionID:   idGenerator.New(),
+			ComputationUsed: 20,
+			Error:           nil,
+			Logs:            []string{},
+			Events:          []flow.Event{},
 		}
 		assert.True(t, trSucceed.Succeeded())
 		assert.False(t, trSucceed.Reverted())
 
 		trReverted := &types.TransactionResult{
-			TransactionID:      idGenerator.New(),
-			ComputationGasUsed: 20,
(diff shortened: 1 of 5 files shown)


sideninja:
Thank you for adding this, it looks good, however, it does highlight the lack of tests for `convert/vm.go` and `types/result.go` so if you could please add those it would be great.

sideninja (changes requested):
Missing tests needs to be implemented

avcdsld:
Thanks for checking the code.
I've added tests for `convert/vm.go` and `types/result.go`. I hope you can check it.

types/result_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+		trSucceed := &types.TransactionResult{
+			TransactionID:      idGenerator.New(),
+			ComputationGasUsed: 20,
+			Error:              nil,
+			Logs:               []string{},
+			Events:             []flow.Event{},
+		}
+		assert.True(t, trSucceed.Succeeded())
+
```
sideninja, line 45: here we could also test for Reverted() being false, and the same on all others.

avcdsld: you are right. added assertions 💪

janezpodhostnik (commented):
Looks great, just added some minor comments about naming.

avcdsld:
> Looks great, just added some minor comments about naming.

I see, I also think that if it is more appropriate to use that term than Gas, it should be modified.
Thanks for the confirmation and the suggestion!

Issue #85, opened 2021-09-22
Default values for service key flags are not working


sideninja:
### Problem
When running the emulator with a specified service private key using the `--service-priv-key` flag, the signature and hashing algo are requested even though they should have default values.

Error:
```
Must specify service key signature algorithm (e.g. --service-sig-algo=ECDSA_P256)
```

### Steps to Reproduce
Run emulator with only `--service-priv-key` value set.

### Acceptance Criteria
It should use default values for sig and hash algo flags.

bluesign:
I think this is already fixed and can be closed.

PR #88, opened 2021-09-29
Emulator Feature: Bootstrap with default contracts

mwufi:
Closes: FLIP Fest issue 27 https://github.com/onflow/flip-fest/issues/27

## Description

The emulator should be able to deploy contracts when it starts up. We add the following ones:
- NFT Marketplace
- NFT
- FUSD

To get this to work, we do two things:
1. make the emulator aware of the service account's private key -- this allows us to sign transactions!
2. Add the default contracts

Notes:
  - For NonFungibleToken, we depend on the `flow-nft` repo
  - For NFTMarketplace, we add the contract directly -- In the future, we can restructure the `nft-marketplace` repo so that it's easier to pull from

We also include defaults for sig&hash algo (in `start.go`). (Does this address https://github.com/onflow/flow-emulator/issues/85 ?)
```
ServiceKeySigAlgo       string        `default:"ECDSA_P256" flag:"service-sig-algo" info:"service account key signature algorithm"`
ServiceKeyHashAlgo      string        `default:"SHA3_256" flag:"service-hash-algo" info:"service account key hash algorithm"`
```

## Tests
`GO111MODULE=on go run ./cmd/emulator --contracts`
image

`GO111MODULE=on go run ./cmd/emulator --contracts --service-priv-key 92ecc6ca2a30b1da782dc6ca6a6da73ad3d9214adb041b5185ca5196bd46e3c7`
If you run with a custom private key, it reflects that:
image

______

For contributor use:


mwufi:
hello @janezpodhostnik @psiemens @sideninja @turbolent :) tagging just in case you didn't see!

cmd/emulator/start/start.go
```diff
@@ -60,6 +60,7 @@ type Config struct {
 	TransactionFeesEnabled bool          `default:"false" flag:"transaction-fees" info:"enable transaction fees"`
 	TransactionMaxGasLimit int           `default:"9999" flag:"transaction-max-gas-limit" info:"maximum gas limit for transactions"`
 	ScriptGasLimit         int           `default:"100000" flag:"script-gas-limit" info:"gas limit for scripts"`
+	BasicContractsOnly     bool          `default:"false" flag:"basic-contracts-only" info:"deploy only basic Flow contracts when emulator starts"`
```
janezpodhostnik, line 63: This can be up for discussion, but I feel the default should be true.

My reasoning is that this will require no change from anyone that is already using the emulator and is depending on the emulator being in a certain state when it starts.

server/contracts.go
```diff
@@ -0,0 +1,63 @@
...
+	"io/ioutil"
+	"regexp"
+
+	emulator "github.com/onflow/flow-emulator"
+	"github.com/onflow/flow-go-sdk"
+	"github.com/onflow/flow-go-sdk/templates"
+	"github.com/onflow/flow-nft/lib/go/contracts"
+)
+
+var ftAddress = flow.HexToAddress("ee82856bf20e2aa6")
```
janezpodhostnik, line 14: I would try not to hardcode this. Can you try using this instead:  https://github.com/onflow/flow-go/blob/3932592e17a289880b693e4e57b9044d09dc194a/fvm/bootstrap.go#L927

server/contracts/NFTStorefront.cdc
```diff
@@ -0,0 +1,495 @@
+import FungibleToken from "./FungibleToken.cdc"
```
janezpodhostnik, line 1: It would be cool if we could also import this one through a go module like we do the `NonFungibleToken` one 🤔.

That would also mean that the emulator doesn't need the import address replacement logic.

janezpodhostnik (commented):
Looks very good overall. Thank you for this!

I left some comments.

cmd/emulator/start/start.go
mwufi: @srinjoyc what do you think? Should we make it an "--include-nft-contracts" flag?

server/contracts.go
mwufi: Ah! Yes, good idea. Will change to be the fvm address

server/contracts/NFTStorefront.cdc
mwufi: Yeah, it would!! But maybe later? It'd just require making a few changes to the `nft-marketplace` repo, to make it friendlier to use as a module. It'd be really neat to make some kind of interface for structuring new contract repos so that they can be imported easily!! (Maybe like how fvm imports the FungibleToken and other contracts!)

On the plus side, by including this file it shows how easy it is to add your own contracts ;)

cmd/emulator/start/start.go
sideninja: I second @janezpodhostnik comment. Although I believe it could be deployed by default we need to be mindful of other devs depending on the current emulator state.

blockchain.go
```diff
@@ -193,17 +193,19 @@ var defaultConfig = func() config {
 // Option is a function applying a change to the emulator config.
 type Option func(*config)
 
-// WithServicePublicKey sets the service key from a public key.
-func WithServicePublicKey(
```
sideninja, line 197: I have doubts if this API should be removed. I think it would be better if `WithServicePrivateKey` would be added, and that one shouldn't have `servicePublicKey` property as it can be inferred from the private key. This is especially true or cases where you don't need these contracts so the private key is not necessary.

blockchain.go
```diff
@@ -193,17 +193,19 @@ var defaultConfig = func() config {
...
+	servicePrivateKey sdkcrypto.PrivateKey,
 	sigAlgo sdkcrypto.SignatureAlgorithm,
 	hashAlgo sdkcrypto.HashAlgorithm,
 ) Option {
 	return func(c *config) {
 		c.ServiceKey = ServiceKey{
-			PublicKey: servicePublicKey,
-			SigAlgo:   sigAlgo,
-			HashAlgo:  hashAlgo,
+			PublicKey:  servicePublicKey,
```
sideninja, line 205: Passing both public and private keys should be avoided, public keys should be inferred from the private keys.

server/contracts.go
```diff
@@ -0,0 +1,63 @@
...
+	deployContract("NFTStorefront", nftStorefront, b, addresses)
+
+	return addresses
+}
+
+func loadContract(name string, replacements map[string]flow.Address) []byte {
+	code := string(readFile(baseContractsPath + name))
+	for name, realAddress := range replacements {
+		placeholder := regexp.MustCompile(fmt.Sprintf(`"[^"\s].*/%s.cdc"`, name))
+		code = placeholder.ReplaceAllString(code, "0x"+realAddress.String())
```
sideninja, line 38: Prefer using `fmt.Sprintf` instead of string concatenation with `+` operator.

server/contracts.go
```diff
@@ -0,0 +1,63 @@
...
+		"FungibleToken":    ftAddress,
+		"NonFungibleToken": nftAddress,
+	})
+	deployContract("NFTStorefront", nftStorefront, b, addresses)
+
+	return addresses
+}
+
+func loadContract(name string, replacements map[string]flow.Address) []byte {
+	code := string(readFile(baseContractsPath + name))
```
sideninja, line 35: I believe it would be better to use filepath package join function https://pkg.go.dev/path/filepath#Join

server/server.go
```diff
@@ -220,7 +232,7 @@ func configureBlockchain(conf *Config, store storage.Store) (*emulator.Blockchai
 	if conf.ServicePublicKey != nil {
 		options = append(
 			options,
-			emulator.WithServicePublicKey(conf.ServicePublicKey, conf.ServiceKeySigAlgo, conf.ServiceKeyHashAlgo),
+			emulator.WithServiceKey(conf.ServicePublicKey, conf.ServicePrivateKey, conf.ServiceKeySigAlgo, conf.ServiceKeyHashAlgo),
```
sideninja, line 235: what if they only pass public key and not private key but use the option to deploy advanced contracts. This edge case should be handled.

server/server.go
```diff
@@ -119,14 +122,23 @@ func NewEmulatorServer(logger *logrus.Logger, conf *Config) *EmulatorServer {
...
 	}
-	logger.WithFields(contracts).Infof("📜  Flow contracts")
+	for contract, address := range contracts {
+		logger.WithFields(logrus.Fields{contract: address}).Infof("📜  Flow contract")
+	}
+
+	if !conf.BasicContractsOnly {
+		addresses := deployContracts(conf, blockchain)
+		for contract, address := range addresses {
+			logger.WithFields(logrus.Fields{contract: "0x" + address.Hex()}).Infof("📜  flow-nft")
```
sideninja, line 139: Use `fmt.Sprintf` instead.

server/contracts.go
```diff
@@ -0,0 +1,63 @@
...
+var ftAddress = flow.HexToAddress("ee82856bf20e2aa6")
+var baseContractsPath = "./server/contracts/"
+
+// deployContracts is meant to be a place where you can load up existing contracts
+// NFTStorefront is an example of how you can do this
+func deployContracts(conf *Config, b *emulator.Blockchain) map[string]flow.Address {
+	addresses := make(map[string]flow.Address)
+
+	nftAddress, _ := deployContract("NonFungibleToken", contracts.NonFungibleToken(), b, addresses)
+	deployContract("ExampleNFT", contracts.ExampleNFT(nftAddress.Hex()), b, addresses)
```
sideninja, line 23: This should be explicitly ignored if not used.

sideninja:
@mwufi Great effort so far, left some comments. Another general comment is don't forget to update docs with any new flags you add.

blockchain.go
mwufi: great. I put this back. it turns out that i don't need to do this -- in fact, in regular usage, this function shouldn't be used at all (since we do want to keep the default service private key :)

sideninja: Exactly, that was worrisome to me.

server/server.go
```diff
@@ -217,7 +230,7 @@ func configureBlockchain(conf *Config, store storage.Store) (*emulator.Blockchai
 		emulator.WithTransactionFeesEnabled(conf.TransactionFeesEnabled),
 	}
 
-	if conf.ServicePublicKey != nil {
+	if conf.ServicePrivateKey == nil && conf.ServicePublicKey != nil {
```
sideninja, line 233: Can you help me understand why do you need to change any functionality regarding passing the private keys in order to finish this task? You deploy contracts by using create account functionality which is working with the current master implementation. I'm not saying you shouldn't improve something if you know how just trying to understand better why you want to change this?

sideninja:
Left a question. Also please make sure the CI tests are passing. By taking a quick look it might be you need to tidy the dependencies.

mwufi:
@sideninja i `go mod tidy`'d the merge conflicts! good to go!

sideninja:
> Hey @psiemens can you quickly review [onflow/fusd#2](https://github.com/onflow/fusd/pull/2)? Made a tiny change to merge before adding it to this one!

I will take a look at this @mwufi

mwufi:
Awesome, thanks @sideninja !!! (other team members, feel free to merge, since i don't have merge permissions!)

server/contracts.go
```diff
@@ -0,0 +1,76 @@
...
+func loadContract(name string, replacements map[string]flow.Address) []byte {
+	code := string(readFile(filepath.Join(baseContractsPath, name)))
+	for name, realAddress := range replacements {
+		placeholder := regexp.MustCompile(fmt.Sprintf(`"[^"\s].*/%s.cdc"`, name))
+		code = placeholder.ReplaceAllString(code, fmt.Sprintf("0x%s", realAddress.String()))
+	}
+	return []byte(code)
+}
+
+func deployContract(name string, contract []byte, b *emulator.Blockchain) (flow.Address, error) {
```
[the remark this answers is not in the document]
mwufi: ah, interesting!! How do you deploy all at once?

Haha I thought it was all right to have different accounts for each one, but they could also go ito the same account... the bigger thing will probably be.. how do you control the accounts once you've deployed the contracts?
- actually it might be a good idea to make them all deploy from the service acct, since that's what you have the keys for, right?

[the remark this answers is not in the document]
sideninja: Yes, I think that it would be better to deploy on the service account not just because of the keys but also there are for sure a lot of tests/projects, unfortunately, depending on the sequence of addresses used for testing so if you will "use" those addresses to do these deployments we might break something for them. You can deploy them using a single transaction as contracts can be passed as an array, see here: https://pkg.go.dev/github.com/onflow/flow-go-sdk/templates#CreateAccount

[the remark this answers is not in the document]
mwufi: done!

image

sideninja (approved):
Great improvement. I think this is it 🎉

mwufi:
wait, actually i made a mistake!! with the second way, it didn't actually submit the transactions, just created them :( so now i'm figuring out how to run the transactions !!

alternatively, would it be a good idea to create one (1) new account for the contracts? instead of deploying to the service acct?

image

sideninja:
> wait, actually i made a mistake!! with the second way, it didn't actually submit the transactions, just created them :( so now i'm figuring out how to run the transactions !!
> 
> alternatively, would it be a good idea to create one (1) new account for the contracts? instead of deploying to the service acct?
> 
> image

I planned to test this tomorrow, so will keep you updated. We would only merge this after the Flip fest so no worries. 
I would advise you to deploy on service account since you have the keys and for the reasons explained before, people assume second address is something they have always got as a second address and now this would shift that.

server/contracts.go
```diff
@@ -0,0 +1,70 @@
...
+		})
+	}
+
+	addresses := []DeployDescription{
+		{"FUSD", serviceAcct, "💵  FUSD contract"},
+		{"NonFungibleToken", serviceAcct, "✨   NFT contract"},
+		{"ExampleNFT", serviceAcct, "✨   NFT contract"},
+		{"NFTStorefront", serviceAcct, "✨   NFT contract"},
+	}
+
```
sideninja, line 51: Use emulator instance `SendTransaction` function here.

mwufi: oh!!! will give that a try -- right now i'm just using the raw blockchain methods :P

image

sideninja: You can just use `CreateAccount()` method on blockchain and avoid handling this yourself.

sideninja: Yeah you are right sorry, no method like that on the blockchain.

mwufi: Ah, okay! I think this'll do it (now it actually checks the service account before printing, to make sure!)

image

PR #90, opened 2021-10-18
Launch local dev wallet instance from emulator

bluesign:
## Description

Emulator launches a local dev wallet instance. 

For Flip Fest: https://github.com/onflow/flip-fest/issues/11

I am planning documentation update after this, but not sure about how to unit test this. 
______

For contributor use:

- [+] Targeted PR against `master` branch


sideninja:
As far as unit testing goes, you can use `httptest` https://pkg.go.dev/net/http/httptest to test routing and serving of couple of files, there's nothing much besides that I believe that needs testing.

sideninja:
@bluesign try to resolve conflicts so this can be marked as done. Also please add a documentation explaining how to run this, you can add it to the readme.

sideninja:
Last couple of small comments I have:
- The flags support env variables by default with snake case, in your case `DEV_WALLET` and `WALLET_PORT` 
- I would change `--wallet-port` to `--dev-wallet-port` just to be sure it's clear and consistent, I always rather error on length than clarity
- Do you think we should change default port for dev-wallet from 3000 to 8701, which is the default port used by dev-wallet and the 3000 is quite popular for local dev of apps. 

These are all quite minor things so even if ignored it's a valid submission.
Great work!

bluesign:
@sideninja thanks, I am updating now. any idea why CI giving error at check-tidy?

sideninja:
> @sideninja thanks, I am updating now. any idea why CI giving error at check-tidy?

did you run tidy again? This kind of CI problems are usually related to go.sum

PR #92, opened 2021-11-03
.git layer and API for snapshots ( for --persist )

bluesign:
## Description

I had to do a second PR as GitHub didn't allow me to change source branch somehow. 

**FLIP Fest issue: CLI Feature: State management #22**

Description

This PR is for issue #22

- Adds .git compatibilty to persistence storage
- Adds emulator HTTP API endpoint
- Implements /emulator/snapshot/{name} command for state management
- Implements /emulator/newBlock command to trigger new Blocks


server/backend/backend.go
```diff
@@ -47,6 +47,11 @@ type Backend struct {
 	automine bool
 }
 
+//hotswap emulator for state management
```
sideninja, line 50:
```suggestion
// hotswap emulator for state management
```

server/emulatorApi.go
```diff
@@ -0,0 +1,85 @@
...
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		m.backend.SetEmulator(blockchain)
+
+		w.WriteHeader(http.StatusOK)
+
+	default:
+		w.WriteHeader(http.StatusInternalServerError)
```
sideninja, line 82: Maybe use `m.server.logger.Error()` to log "State management only available with badger storage"

bluesign:
@sideninja I stuck in those conflicts, this time tidy didn't help, any idea here?

sideninja:
I was testing the state management API and one thing that I would change is logging. Depending on the verbosity of logs you either see the first logs (non-verbose) or second (verbose). I feel that the first logs are too verbose for non-verbose mode. I also think both are missing a log that would say the state was either created or changed, so for example when I would create a new state I would get a log in sense of `Created a new state snapshot with the name "test-1"` and then when switching between states the log `Switched to snapshot with name "test-2"`. This log would be the only one in a non-verbose log as the current logs are quite low-level logs that would only confuse developers. Even in verbose mode I would only keep the most critical logs, I don't think any of these logs are critical to understanding what is happening or there are some that we should keep but only in verbose mode.
Non-Verbose
```
INFO[0004] Got compaction priority: {level:0 score:1.73 dropPrefix:[]} 
INFO[0004] All 1 tables opened in 0s                    
INFO[0004] Replaying file id: 0 at offset: 299733       
INFO[0004] Replay took: 2.041µs 
```
Verbose
```
DEBU[0238] Storing value log head: {Fid:0 Len:29 Offset:299704} 
INFO[0238] Got compaction priority: {level:0 score:1.73 dropPrefix:[]} 
INFO[0238] Running for level: 0                         
DEBU[0238] LOG Compact. Added 242 keys. Skipped 8 keys. Iteration took: 6.844375ms 
DEBU[0238] Discard stats: map[]                         
INFO[0238] LOG Compact 0->1, del 2 tables, add 1 tables, took 45.617458ms 
INFO[0238] Compaction for level: 0 DONE                 
INFO[0238] Force compaction on level 0 done             
INFO[0238] All 1 tables opened in 0s                    
INFO[0238] Replaying file id: 0 at offset: 299733       
INFO[0238] Replay took: 1.667µs                         
DEBU[0238] Value log discard stats empty   
```

Another thing that I feel could be useful is for API to have some sort of response body, now you indicate success with status which is great, but it would be nice to have some response back weather a new state was created or weather the emualtor switch to that state, I could even argue it would be better to have creation of state limited to POST method and switching to PUT but I'm totally fine if it stays this way as long as there is some response so the client can now if the state was created or switched to.

As far as the code goes it's very well written, I just left some more aesthetic comments

server/emulatorApi.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (m EmulatorApiServer) Snapshot(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	vars := mux.Vars(r)
+	name := vars["name"]
+
+	switch (*m.storage).Store().(type) {
+	case *badger.Store:
+		badgerStore := (*m.storage).Store().(*badger.Store)
+		badgerStore.JumpToContext(name)
```
sideninja, line 68: Explicity ignore error if you are sure there is no need to handle it

storage/badger/store.go
```diff
@@ -21,43 +21,236 @@ package badger
...
 
 var _ storage.Store = &Store{}
 
+func getTag(r *git.Repository, tag string) *object.Tag {
+	tags, err := r.TagObjects()
+	if err != nil {
+		return nil
+	}
+	var res *object.Tag = nil
+	tags.ForEach(func(t *object.Tag) error {
```
sideninja, line 59: Explicity ignore error if you are sure there is no need to handle it

storage/badger/store.go
```diff
@@ -21,43 +21,236 @@ package badger
...
+		Name:  name,
+		Email: email,
+		When:  time.Now(),
+	}
+}
+
+//prevents git commits when emulator running
+func (s *Store) lockGit() {
+	lockPath := fmt.Sprintf("%s/.git/index.lock", s.path)
+	ioutil.WriteFile(lockPath, []byte("emulatorLock"), 0755)
```
sideninja, line 96: Explicity ignore error if you are sure there is no need to handle it

storage/badger/store.go
```diff
@@ -21,43 +21,236 @@ package badger
...
+}
+
+func (s *Store) unlockGit() {
+	lockPath := fmt.Sprintf("%s/.git/index.lock", s.path)
+	os.Remove(lockPath)
```
sideninja, line 101: Explicity ignore error if you are sure there is no need to handle it

storage/badger/store.go
```diff
@@ -21,43 +21,236 @@ package badger
...
+		created, err := setTag(s.dbGitRepository, context, defaultSignature("Emulator", "emulator@onflow.org"))
+		if err != nil && !created {
+			return err
+		}
+
+	} else {
+
+		//create new branch
+		uuidWithHyphen := uuid.New()
+		uuid := strings.Replace(uuidWithHyphen.String(), "-", "", -1)
```
sideninja, line 139: conflicts with imported package

storage/badger/store.go
```diff
@@ -21,43 +21,236 @@ package badger
...
+		err := w.Checkout(&git.CheckoutOptions{Create: true, Force: true, Branch: plumbing.NewBranchReferenceName(uuid)})
+		if err != nil {
+			return err
+		}
+
+		//we have new branch but we don't need to create a tag, we just need to reset to tag
+		tag := getTag(s.dbGitRepository, context)
+		if tag != nil && !tag.Hash.IsZero() {
+			commit, _ := tag.Commit()
+			w.Reset(&git.ResetOptions{
```
sideninja, line 150: Explicity ignore error if you are sure there is no need to handle it

storage/badger/store.go
```diff
@@ -21,43 +21,236 @@ package badger
...
+	}
+
+	return nil
+
+}
+
+func (s *Store) newCommit(message string) {
+	s.unlockGit()
+	defer s.lockGit()
+	s.Sync()
```
sideninja, line 170: Explicity ignore error if you are sure there is no need to handle it

storage/badger/store.go
```diff
@@ -21,43 +21,236 @@ package badger
...
+		if filepath.Ext(path) == ".vlog" || filepath.Ext(path) == ".sst" {
+			w.Add(path[strings.LastIndex(path, "/")+1:])
+		}
+		return nil
+	})
+	if err != nil {
+		panic(err)
+	}
+
+	w.Commit(message, &git.CommitOptions{
```
sideninja, line 191: Explicity ignore error if you are sure there is no need to handle it

storage/badger/store.go
```diff
@@ -21,43 +21,236 @@ package badger
...
 // New returns a new Badger Store.
 func New(opts ...Opt) (*Store, error) {
 	badgerOptions := getBadgerOptions(opts...)
-
+	badgerOptions.BypassLockGuard = true
 	db, err := badger.Open(badgerOptions)
 	if err != nil {
 		return nil, fmt.Errorf("could not open database: %w", err)
 	}
+	db.Sync()
```
sideninja, line 224: Explicity ignore error if you are sure there is no need to handle it

storage/badger/store.go
```diff
@@ -225,13 +420,37 @@ func (s *Store) CommitBlock(
...
+
+			message = fmt.Sprintf("%sArguments (%d): \n\n", message, len(tx.Arguments))
+			for argID, arg := range tx.Arguments {
+				message = fmt.Sprintf("%s\t- Argument %d: %s\n", message, argID, string(arg))
+			}
+
+			message = fmt.Sprintf("%sCode:\n%s", message, tx.Script)
+
+			result := transactionResults[txID]
+			insertTransactionResult(txID, *result)(txn)
```
sideninja, line 434: forgot err probably, as bellow err is always nil

sideninja:
This is really awesome log
```
Author: Flow Emulator <emulator@onflow.org>
Date:   Fri Nov 5 15:18:38 2021 +0100

    Committed Block: 06cb1d0d3446335b20e3e045ea9896a305ba90f69c42bbe456a17bb3513e6dc8
    Transaction    : 63737abf8bd5aa66af64d98f2e12405151293d4dfc2ac1f0379de27bc397fd74
    
    Arguments (2):
    
            - Argument 0: {"type":"Array","value":[]}
    
            - Argument 1: {"type":"Dictionary","value":[]}
    
    Code:
    
    transaction(publicKeys: [String], contracts: {String: String}) {
            prepare(signer: AuthAccount) {
                    let acct = AuthAccount(payer: signer)
    
                    for key in publicKeys {
                            acct.addPublicKey(key.decodeHex())
                    }
    
                    for contract in contracts.keys {
                            acct.contracts.add(name: contract, code: contracts[contract]!.decodeHex())
                    }
            }
    }
    Result:
    
            - Error Message : [1006] [Error Code: 1006] invalid proposal key: public key 0 on account f8d6e0586b0a20c7 does not have a valid signature: [Error Code: 1009] invalid envelope key: public key 0 on account f8d6e0586b0a20c7 does not have a valid signature: signature is not valid
    
            - Logs (0):
    
            - Events (0):
       
```

sideninja:
> @sideninja I stuck in those conflicts, this time tidy didn't help, any idea here?

I will help you after the last commit.

storage/badger/store.go
bluesign: can you explain this a bit more?

sideninja: just the fact that uuid is name of imported package, and it's better idea to name a variable differently to not conflict the namespace.

bluesign: ah ok thanks

sideninja (approved):

PR #132, opened 2022-02-21
fix segfault on already listening on port

bluesign:


sideninja:
Awesome! Nice catch. Just wondering tho, if the emulator ports are in use should we stop the process and return a non-zero exit code to indicate that we indeed didn't accomplish the desired result? I think this would be beneficial to the emulator being integrated into bash scripts that need to have a way to see if the emulator startup was a success. What do you think? But in any case really nice this has been handled, we just need to be aware that the launch procedure is now different and there might be some environments/CI detecting successful startup by the error code.

sideninja (approved):

PR #164, opened 2022-06-18
Start implementation of a debugger

turbolent:
## Description

Start the implementation of a debugger, by implementing a server which speaks the [Debug Adapter Protocol](https://microsoft.github.io/debug-adapter-protocol/). 

Implement enough of the protocol to:
- Allow a script to get executed and report the result or error to the client
- Allow execution to be paused and continued
- Step through a program
- Provide sources for scripts and contracts
- Provide the stack trace
- Evaluate variables
- Add breakpoints
- Show variables in scope

https://user-images.githubusercontent.com/51661/174458756-5adc1dd0-cb5e-4805-aa8f-df8b896bc85d.mov

https://user-images.githubusercontent.com/51661/174458757-3534acc0-f684-4b03-ad2f-b9ae00272616.mov

https://user-images.githubusercontent.com/51661/174493185-bdfa8fc0-f773-414c-851a-3a0790c66287.mov

______

For contributor use:


bluesign:
This is amazing, thank you very much.

janezpodhostnik:
This is so amazing ❤️ 🔥

turbolent:
This PR currently depends on multiple feature and bugfix PRs in other repos, I'll update it to proper released versions once they're available

sideninja (approved):
Awesome. The only thing missing would be to update the docs.

server/debugsession.go
```diff
@@ -0,0 +1,578 @@
...
+	case *dap.VariablesRequest:
+		// TODO: reply with error if ds.stop == nil
+
+		inter := ds.stop.Interpreter
+
+		activation := ds.debugger.CurrentActivation(inter)
+
+		functionValues := activation.FunctionValues()
+
+		variables := make([]dap.Variable, 0, len(functionValues))
```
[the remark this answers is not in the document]
turbolent: Yes, that could be added too

turbolent:
Merged master and noticed that I accidentally removed some required functionality from Cadence in its latest release. 
Brought it back in https://github.com/onflow/cadence/pull/1924, so this PR is temporarily depending on that PR until it is released.

server/debugsession.go
```diff
@@ -0,0 +1,578 @@
...
+			Body: dap.StackTraceResponseBody{
+				StackFrames: stackFrames,
+			},
+		})
+
+	case *dap.SourceRequest:
+		path := request.Arguments.Source.Path
+
+		code := ds.pathCode(path)
+
```
bluesign, line 253: I think here can be better to get codes from CadenceDebugger. So we can support also `Attach`.

server/debugsession.go
```diff
@@ -0,0 +1,578 @@
...
+	case *dap.PauseRequest:
+		ds.debugger.RequestPause()
+
+		ds.send(&dap.PauseResponse{
+			Response: newDAPSuccessResponse(request.GetRequest()),
+		})
+
+	case *dap.NextRequest:
+		ds.step()
```
bluesign, line 215: I think this is step into, but comment on L222 makes it seem this is step-over.

bluesign:
I think supporting `Attach` can be great for usage too. So we can offload most of the transaction and script arguments etc to CLI.

sideninja:
I just changed it to draft since there's no activity. 
Should we get this in tho @turbolent I really like the functionality but don't know what are your thoughts about this and what you think we should do additionally.

turbolent:
Thanks @sideninja!

I would still love to get it in, but:
- I would like another review
- After the last FVM refactoring, it is not clear to me how to restore the code that sets a debugger on the runtime

@janezpodhostnik @pattyshack Could you maybe please help here?

server/debugsession.go
turbolent: Agreed, it would be nice to request this from Cadence, but the Cadence debugger and interpreter do not have the source code (only the AST).

server/debugsession.go
turbolent: Yes, all stepping requests are just simple steps, there is no support for step into, step over, step out etc.

pattyshack:
could you give me a bit more detail?  How was the debugger set prior to the refactoring?

turbolent:
@pattyshack Previously the runtime was exposed in this way: https://github.com/onflow/flow-emulator/pull/164/files#diff-1fb2891d150a66e1319495e4f10a7ce04855476836a3a6b17f253e312f37d23bR1362-R1363

pattyshack:
you can provide custom runtimes to the vm using NewCustomReusableCadenceRuntimePool and WithReusableCadenceRuntimePool

see:
https://github.com/onflow/flow-go/blob/master/fvm/runtime/reusable_cadence_runtime.go#L203
https://github.com/onflow/flow-go/blob/master/fvm/context.go#L344

example:
https://github.com/onflow/flow-go/blob/master/engine/execution/computation/computer/computer_test.go#L428

bluesign:
@turbolent if it is ok with you, can I take this over?

turbolent:
@bluesign of course, please do!

sideninja:
> @bluesign of course, please do!

> @turbolent if it is ok with you, can I take this over?

I was just working on it Friday and plan to continue today, I have a branch in progress.

turbolent:
Continued in #294

PR #165, opened 2022-06-23
Organize content for docs refresh caching

codingone21:
Closes https://github.com/onflow/next-docs-v1/issues/201

## Description
- Copied readme into the /docs content folder
______

For contributor use:

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -1,0 +1,167 @@
+<br />
+<p align="center">
+  <a href="https://docs.onflow.org/emulator/">
+    <img src="./emulator-banner.svg" alt="Logo" width="410" height="auto">
+  </a>
+
+  <p align="center">
+    <i>The Flow Emulator is a lightweight tool that emulates the behaviour of the real Flow network.</i>
+    <br />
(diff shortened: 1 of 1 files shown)


sideninja:
Is it possible to just include the main README? I think we should transition to an emulator having more docs which would be hosted in `/docs` but until then I worry we have to keep two copies identical. If it's not possible (or it proves to be difficult) I would say continue with this approach.

codingone21:
> 

That's true, it is possible to use the README it it's the only file being used.  I briefly discussed it with Mack and we thought it would be better to keep the path consistent for the docs site. Also, if we decide to add more content to the docsite, eventually the /docs content would be structured differently from README. I'm open to changes, and an alternative could be using Github Actions to copy any changed README to the /docs/overview file. This way, we don't need to update both manually.

sideninja:
> docs content would be structured differently from README. I'm open to changes, and an alternative could be using Github Actions to copy any changed README to the /docs/overview file. This way, we don't need to update both manually.

Good idea.

PR #166, opened 2022-06-23
custom zerolog start

bjartek:
Adds support for sending in a custom zerolog to emulator to get debug information

### DIFF
--- a/transaction_test.go
+++ b/transaction_test.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"strings"
 	"testing"
 
@@ -1682,7 +1683,7 @@ func TestSubmitTransactionWithCustomLogger(t *testing.T) {
 
 	t.Parallel()
 	var memlog bytes.Buffer
-	memlogWrite := bufio.NewWriter(&memlog)
+	memlogWrite := io.Writer(&memlog)
 	logger := zerolog.New(memlogWrite).Level(zerolog.DebugLevel)
 
 	b, err := emulator.NewBlockchain(
--- a/transaction_test.go
+++ b/transaction_test.go
@@ -1734,19 +1734,16 @@ func TestSubmitTransactionWithCustomLogger(t *testing.T) {
 	}
 
 	assert.NoError(t, err)
-	assert.Equal(t, uint(1), meter.ComputationIntensities[common.ComputationKindLoop])
-	assert.Equal(t, 8, meter.ComputationUsed)
+	assert.Equal(t, `{LedgerInteractionUsed:9373 ComputationUsed:18 MemoryUsed:0 ComputationIntensities:map[Statement:10 Loop:1 FunctionInvocation:8 CreateCompositeValue:2 TransferCompositeValue:3 CreateArrayValue:4 TransferArrayValue:446 CreateDictionaryValue:1 ComputationKind(2005):1 ComputationKind(2006):1 ComputationKind(2007):1 ComputationKind(2008):1 ComputationKind(2011):2 ComputationKind(2015):2 ComputationKind(2017):2 ComputationKind(2020):4760 ComputationKind(2022):1 ComputationKind(2025):3 ComputationKind(2026):1621 ComputationKind(2027):1] MemoryIntensities:map[]}`, fmt.Sprintf("%+v", meter))
+
 }
 
 type Meter struct {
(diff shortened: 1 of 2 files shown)


bjartek:
would it make sense to have an option to the emulator to actually set a custom logger, parse it and return the debug information? The reason I am doing this PR is because I want to expose this as data in overflow

```
{"level":"info","txHash":"232553ad083557f861850316704d6d5295cbc73fe6ebc0dfa666b42dd9be1c4f","ledgerInteractionUsed":174402,"computationUsed":8,"memoryUsed":0,"computationIntensities":{"2027":1,"2026":150,"1003":3,"2006":1,"2007":2,"1026":446,"1025":4,"1001":4,"2022":1,"2025":1,"1040":1,"1002":1,"2005":2,"2020":4268,"2011":1,"2015":1,"1010":1},"memoryIntensities":{},"message":"transaction execution data"}
```

sideninja:
I think this is a great addition for sure. I wonder do you think maybe we could by default pass in the logger configured on the backend, but it would need to log FVM logs under a `trace` level so it doesn't spam the default output when running the emulator. 

> would it make sense to have an option to the emulator to actually set a custom logger, parse it and return the debug information? The reason I am doing this PR is because I want to expose this as data in overflow

So like accepting an interface which would then be built by unmarshaling JSON log into it? I feel that might not fall under the emulator domain to be honest. Maybe we could start without it and if there's a common usage of it we could add it later, but I wouldn't want to tailor the emulator too much towards a specific use case if that makes sense?

bjartek:
> but it would need to log FVM logs under a trace level so it doesn't spam the default output when running the emulator.
The problem with that is Meters are only logged if debug is enabled. Would it be possible to change that one aswell?

Honestly I think that the default should just print to stdout. I wrote some code in overflow to write to buffer and marshal the output for assertions in tests.

sideninja:
> Honestly I think that the default should just print to stderr. I wrote some code in overflow to write to buffer and marshal the output for assertions in tests.

Not sure if stderr would be appropriate. It's good you have the logger and then you would just specify the verbosity of output, based on that the logger would print, but in your case you wouldn't print to output anyway since you will be directly consuming logger.

sideninja (approved):

bjartek:
@janezpodhostnik I added a test that compares a Sprtinf("%+v") of the meter struct against a string. Not sure if there is a cleaner way to get the named values of the Kinds as using pure numbers here is not very readable to me atleast.

transaction_test.go
```diff
@@ -1672,3 +1678,73 @@ func TestInfiniteTransaction(t *testing.T) {
...
+	scanner := bufio.NewScanner(&memlog)
+	for scanner.Scan() {
+		txt := scanner.Text()
+		if strings.Contains(txt, "transaction execution data") {
+			err = json.Unmarshal([]byte(txt), &meter)
+		}
+	}
+
+	assert.NoError(t, err)
+	assert.Equal(t, `{LedgerInteractionUsed:9373 ComputationUsed:18 MemoryUsed:0 ComputationIntensities:map[Statement:10 Loop:1 FunctionInvocation:8 CreateCompositeValue:2 TransferCompositeValue:3 CreateArrayValue:4 TransferArrayValue:446 CreateDictionaryValue:1 ComputationKind(2005):1 ComputationKind(2006):1 ComputationKind(2007):1 ComputationKind(2008):1 ComputationKind(2011):2 ComputationKind(2015):2 ComputationKind(2017):2 ComputationKind(2020):4760 ComputationKind(2022):1 ComputationKind(2025):3 ComputationKind(2026):1621 ComputationKind(2027):1] MemoryIntensities:map[]}`, fmt.Sprintf("%+v", meter))
```
janezpodhostnik, line 1737: This is a bit strict I think. I had something like this in mind:
```suggestion
assert.Greater(t, meter. LedgerInteractionUsed, 0)
assert.Greater(t, meter. ComputationUsed, 0)
assert.Equal(t, meter. MemoryUsed, 0) // this will change with the new emulator version I think
assert.Greater(t, meter. ComputationIntensities.len, 0)
assert.Equal(t, meter. MemoryIntensities.len, 0) // this will change with the new emulator version I think
```

janezpodhostnik (approved):
Just one more comment otherwise looks good to me.

sideninja:
@bjartek some problems with tests

bjartek:
CI should be fixed now

bjartek:
Status?

PR #174, opened 2022-06-30
[Don't Merge] remove the controller from storage

ramtinms:
Closes #???

## Description

______

For contributor use:

### DIFF
--- a/storage/badger/keys.go
+++ b/storage/badger/keys.go
@@ -100,10 +100,9 @@ func ledgerKey(blockHeight uint64) []byte {
 }
 
 func ledgerChangelogKey(registerID flowgo.RegisterID) []byte {
-	return []byte(fmt.Sprintf("%s-%s-%s-%s",
+	return []byte(fmt.Sprintf("%s-%s-%s",
 		ledgerChangelogKeyPrefix,
 		registerID.Owner,
-		registerID.Controller,
 		registerID.Key))
 }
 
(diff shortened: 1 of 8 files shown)


janezpodhostnik (commented):
This looks good.

The concern i have is that anyone using persistent storage will have to start fresh, as the reads are no longer compatible.

One solution would be to fetch registers at both the old key and the new key, but we would have to keep that code in for a long time. 

The other solution would be to offer some persistent storage migration tool or maybe a guide guide or just automatically migrate persistent storage at first load.

janezpodhostnik (approved):

storage/badger/keys.go
```diff
@@ -115,14 +114,26 @@ func ledgerValueKey(registerID flowgo.RegisterID, blockHeight uint64) []byte {
...
+		parts := strings.SplitN(registerString, "-", 3)
+		if len(parts) < 3 {
+			return flowgo.RegisterID{}, fmt.Errorf("failed to parse register ID from %s", string(key))
+		}
+		return flowgo.RegisterID{
+			Owner: parts[0],
+			Key:   parts[2], // skip the controller
+		}, nil
+	}
+	// else is the new format
```
turbolent, line 129: Do we need this backwards-compatibility. Can we just remove it?

ramtinms: I thought we don't need it initially, but Janez is right people might have use persistant storage that have these values, it seems there is no harm to keep it in case we have to load things.

turbolent: Sounds good!

turbolent (approved):

psiemens:
@ramtinms It looks like this branch references [flow-go commit 7015f054a69f](https://github.com/onflow/flow-go/commit/7015f054a69f) that isn't on `master` (I think it's in this PR: https://github.com/onflow/flow-go/pull/2713).

After you merge that PR, can you update the emulator to import from `flow-go@master`? We're running into downstream co-dependency issues because commit 7015f054a69f doesn't have some of the changes needed from master.

In the future, please avoid merging an emulator PR the depends on unmerged code in flow-go 🙏

ramtinms:
@psiemens good point, going to update this as soon as the other PR is merged, I also going to create an issue so we could remove the dependency of flow-go to flow-emulator, that seems not natural.

ramtinms:
updated here https://github.com/onflow/flow-emulator/pull/177

PR #179, opened 2022-07-17
"Skip transaction validation" option 

### LINKED ISSUE #169: Implement a flag to skip transaction validation
### Issue To Be Solved

Currently in many use cases (playground, testing etc) there is a need to skip some verifications in transactions. Signature, Proposer sequence number etc

### Suggest A Solution

some flag like `--skip-tx-validation` which will allow to bypass signature checks would be really useful.

### Context

I am building a playground backed by emulator. 

cc: @sideninja

bluesign:
Closes #169

### DIFF
--- a/blockchain.go
+++ b/blockchain.go
@@ -127,18 +127,19 @@ func GenerateDefaultServiceKey(
 
 // config is a set of configuration options for an emulated blockchain.
 type config struct {
-	ServiceKey                ServiceKey
-	Store                     storage.Store
-	SimpleAddresses           bool
-	GenesisTokenSupply        cadence.UFix64
-	TransactionMaxGasLimit    uint64
-	ScriptGasLimit            uint64
-	TransactionExpiry         uint
-	StorageLimitEnabled       bool
-	TransactionFeesEnabled    bool
-	MinimumStorageReservation cadence.UFix64
-	StorageMBPerFLOW          cadence.UFix64
-	Logger                    zerolog.Logger
+	ServiceKey                   ServiceKey
+	Store                        storage.Store
+	SimpleAddresses              bool
+	GenesisTokenSupply           cadence.UFix64
+	TransactionMaxGasLimit       uint64
+	ScriptGasLimit               uint64
+	TransactionExpiry            uint
+	StorageLimitEnabled          bool
+	TransactionFeesEnabled       bool
+	MinimumStorageReservation    cadence.UFix64
+	StorageMBPerFLOW             cadence.UFix64
+	Logger                       zerolog.Logger
+	TransactionValidationEnabled bool
 }
 
 func (conf config) GetStore() storage.Store {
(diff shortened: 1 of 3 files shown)


bluesign:
Not sure how to test this logic, I appreciate any assistance.

sideninja:
I was thinking of announcing this kind of feature as experimental at first, what do you think? My thinking is that this makes the execution of transactions act different to how they would work on mainnet and hence I would like to have the user understand that, one way is to accept the experimental mode. Maybe some of the features would require to use additional `---experiemental` flag which in docs would be explained that this is potentially "dangerous" territory but it allows you have way more freedom (of course in nicer words). This is a bit subjective for sure but feels right to me, what do you think @bluesign 
cc @psiemens

bluesign:
@sideninja  `--experiemental` is great idea actually. Or maybe we can prefix flags with `experiemental`.
`--experimental-skip-tx-validation` ( or can be another prefix `advanced`, `expertmode` etc )

blockchain.go
```diff
@@ -332,6 +334,16 @@ func WithTransactionFeesEnabled(enabled bool) Option {
 	}
 }
 
+// WithTransactionValidationEnabled enables/disables transaction valudation.
+//
+// If set to false transactions don't check for signatures or sequence numbers.
+// The default is true.
```
psiemens, line 340:
```suggestion
// WithTransactionValidationEnabled enables/disables transaction validation.
//
// If set to false, the emulator will not verify transaction signatures or validate sequence numbers.
//
// The default is true.
```

cmd/emulator/start/start.go
```diff
@@ -38,31 +38,32 @@ import (
...
+	StorageMBPerFLOW          string        `flag:"storage-per-flow" info:"the MB amount of storage capacity an account has per 1 FLOW token it has. e.g. '100.0'. The default is taken from the current version of flow-go"`
+	MinimumAccountBalance     string        `flag:"min-account-balance" info:"The minimum account balance of an account. This is also the cost of creating one account. e.g. '0.001'. The default is taken from the current version of flow-go"`
+	TransactionFeesEnabled    bool          `default:"false" flag:"transaction-fees" info:"enable transaction fees"`
+	TransactionMaxGasLimit    int           `default:"9999" flag:"transaction-max-gas-limit" info:"maximum gas limit for transactions"`
+	ScriptGasLimit            int           `default:"100000" flag:"script-gas-limit" info:"gas limit for scripts"`
+	WithContracts             bool          `default:"false" flag:"contracts" info:"deploy common contracts when emulator starts"`
+	SkipTransactionValidation bool          `default:"false" flag:"skip-tx-validation" info:"skip verification of transaction signature(s) and proposer sequence number"`
```
psiemens, line 66:
```suggestion
	SkipTransactionValidation bool          `default:"false" flag:"skip-tx-validation" info:"skip verification of transaction signatures and sequence numbers"`
```

bluesign:
if good for all:
- I am renaming flag to `--experimental-skip-tx-validation` 
- updating documentation

sideninja:
> However, I think an experimental flag should only be used to gate commands that a) are new and may not work as expected or b) have a high chance of breaking in the future.
I generally prefer using a prefix instead of a separate flag.
IMO the best way to communicate that a flag is an advanced feature is to give it a long and scary name. This feels scary enough to me --skip-tx-validation TBH

Fair enough. I still think having a group command tagged as advanced commands would be good, then again experimental might more indicate instability and breaking stuff and might not be best suited, but maybe "advanced" (like bluesign wrote) might be good, so you could then have `--advanced-skip-tx-validation` and `--advanced-storage-changes` (duuno), but I like the idea of having those commands be grouped. I don't feel strongly about it, it was just an idea. So I'm fine with either direction here, I think the important thing is the cool feature :)

sideninja (approved):

sideninja:
@bluesign can you resolve the conflicts when you have the time so we merge this in.

PR #187, opened 2022-08-10
Upgrade flow to v0.27.1

pattyshack:
Closes #???

## Description

Upgrade flow to v0.27.1

______

For contributor use:


server/backend/backend.go
```diff
@@ -28,7 +28,7 @@ import (
 	jsoncdc "github.com/onflow/cadence/encoding/json"
 	sdk "github.com/onflow/flow-go-sdk"
 	"github.com/onflow/flow-go/access"
-	fvmerrors "github.com/onflow/flow-go/fvm/errors"
+	fvmerrors "github.com/onflow/flow-go/fvm/errors" // nolint:typecheck
```
sideninja, line 31: why do we need this ignored?

pattyshack: the linter is incorrectly reporting this as an error (i didn't want to spent too much time looking into this)

sideninja: if the linter has these bugs shouldn't we rather just use an older version then?

.github/workflows/ci.yml
```diff
@@ -39,11 +39,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.18.x'
       - name: Run golangci-lint
-        uses: golangci/golangci-lint-action@v2.3.0
+        uses: golangci/golangci-lint-action@v3
```
sideninja, line 46: we should make sure the makefile is installing the same golangci linter

server/backend/backend.go
pattyshack: looks like the linter was compiled using the wrong go version.  these are fixed now

.github/workflows/ci.yml
pattyshack: bumped version in make file

sideninja (approved):
Thank you. LGTM

PR #193, opened 2022-08-24
Add --host command line flag

### LINKED ISSUE #192: Emulator triggers firewall warnings on every startup
## Instructions

Prevent flow-emulator from creating a firewall dialog every time it runs.

### Issue To Be Solved

Currently, the flow-emulator listens on 0.0.0.0 which causes a firewall dialog to appear every time the emulator runs.  This can be quite frustrating when the emulator is not run as a standalone, but rather automated by something such as the Flow JS Testing framework.

Listen on 127.0.0.1 loopback address instead of 0.0.0.0.  This solution is only valid if there is no existing purpose for the emulator to accept any incoming connections and only local ones.

If there is a reason to listen for incoming connections, this could be done as a command line flag.

### (Optional): Context

[@onflow/flow-js-testing](https://github.com/onflow/flow-js-testing) spawns a new child process of the flow-emulator every time it runs a test.  This means that when running tests on their local machine, developers will get hit with a new firewall dialog every time a test runs (i.e. on Mac "do you want to accept incoming connections" and likely something similar on Windows)

jribbink:
Closes #192

## Description

This PR makes the servers (rest/http/grpc) listen on the loopback address 127.0.0.1 (local connections) instead of 0.0.0.0 (all incoming connections).  As long as the emulator has no existing needs to listen for inbound connections this would provide a solution to hide firewall dialogs which appear on emulator startup.

If this cannot be done, it would be extremely useful to have a command line argument to listen locally instead and I can make a PR for this if needed.

______

For contributor use:


turbolent:
Nice find! I was always wondering why the dialog popped up. 

Could you please make it a flag that defaults to `127.0.0.1` instead? That would also keep all the places where the magic string is used now in sync.

jribbink:
> Nice find! I was always wondering why the dialog popped up.
> 
> Could you please make it a flag that defaults to `127.0.0.1` instead? That would also keep all the places where the magic string is used now in sync.

Cheers, I've gone ahead and done that.  Let me know if you want anything else changed.

bluesign:
I think ‘host’ flag can be better here, with value ‘localhost’ as default

cmd/emulator/start/start.go
```diff
@@ -65,6 +65,7 @@ type Config struct {
 	ScriptGasLimit            int           `default:"100000" flag:"script-gas-limit" info:"gas limit for scripts"`
 	WithContracts             bool          `default:"false" flag:"contracts" info:"deploy common contracts when emulator starts"`
 	SkipTransactionValidation bool          `default:"false" flag:"skip-tx-validation" info:"skip verification of transaction signatures and sequence numbers"`
+	Host                      string				`default:"0.0.0.0" flag:"host" info:"host to listen on for emulator GRPC/REST/Admin servers"`
```
sideninja, line 68: could we default to 127.0.0.1?

sideninja:
Nice, didn't know that was the reason, but makes sense. I would default to `127.0.0.1`. Also please make sure you also update the readme and docs with this flag.

cmd/emulator/start/start.go
jribbink: for sure, only reason I did 0.0.0.0 was so it wouldn't break any existing functionality of the emulator but this is more sensible.

README.md
```diff
@@ -37,33 +37,36 @@ The Flow Emulator can be run in different modes and settings, all of them are de
...
+| `--with-contracts`              | `FLOW_WITHCONTRACTS`             | `false`        | Deploy common contracts when emulator starts                                                                                                                                                                                                        |
+| `--skip-transaction-validation` | `FLOW_SKIPTRANSACTIONVALIDATION` | `false`        | Skip verification of transaction signatures and sequence numbers                                                                                                                                                                                    |
+| `--host`                        | `FLOW_HOST`                      | `100000`       | Host to listen on for emulator GRPC/REST/Admin servers                                                                                                                                                                                              |
```
sideninja, line 69: I don't think the default value is correct.

sideninja:
Have you manually tested out setting different hosts and checked if servers are working correctly? We don't have a test for this.

sideninja:
Have you manually tested out setting different hosts and checked if servers are working correctly? We don't have a test for this.

jribbink:
> Have you manually tested out setting different hosts and checked if servers are working correctly? We don't have a test for this.

Yup, 127.0.0.1/localhost and 0.0.0.0/empty host behave as expected

sideninja (approved):
Thank you!

PR #220, opened 2022-10-16
Backend cleanup ( adapter removal ) 

bluesign:
Closes #???

Removed the adapter, as it lost its purpose over time. This should increase our test coverage in the long run.  

______

For contributor use:


janezpodhostnik (approved):
I agree that the adapter looks unnecessary, and the code changes look good.

sideninja (approved):
Nice cleanup

sideninja:
@bluesign actually updating this into CLI made me realize that we probably shouldn't change the types of the backend. It's ok to remove the adapter as that's not used anywhere but changing types to flow-go types break any integration of the emulator. Also, I think returning SDK types makes it easier to use. 
Do you have some reasons why you would like to keep them flow-go types?

bluesign:
@sideninja what was the breaking point when you were trying to integrate into CLI ? Somehow CLI uses backend ?

sideninja:
> I am sorry I totally missed this side effect but merging this will move all this backend adapter code to flowkit which doesn't make any sense.

Exactly. And it's me who missed it. 

I agree I think that is some tech debt we could address by cleaning and unifying some models, but honestly, I wouldn't hold my breath as there are other teams I don't think will have time to prioritize this. 

I think we could still remove the adapter part but just keep the backend returning SDK types. 

Flowkit and Go SDK should have uniform models, if you find some that are not I can address that.

bluesign:
@sideninja, my go knowledge is a bit limited, is it possible to make two structs somehow compatible in a 3rd one? 

I mean, something like SDK.address can be also accepted as Flow.address

sideninja:
You can compose structs but the types won't match, the only way to my knowledge of doing anything like it is defining an interface for Address, which arguably could be a very nice way of accepting any of those types since they all implement the basic methods.

bluesign:
Ok I have looked a bit more;

Now seems layers as follows: Emulator > Backend > FlowKit  and Emulator > Backend > BackEndAdapter  

What would you think about: Emulator ( already using flow-go-sdk types )  > FlowKit  and Emulator > Backend ( using flowgo types to be compatible with AN.API ) ?

sideninja:
There is some extra implementation for the events in the backend https://github.com/onflow/flow-emulator/blob/master/server/backend/backend.go#L435-L560 

It might be sensible to move that into the emulator, but I'm not in the right mindset now (would have to dig through code to get in the right context) to really decide that. 

I feel that backend is also a weird naming for it after. Also the logging not sure if it makes sense to be in that layer.

Issue #130, opened 2022-02-16
Support non-local databases


KeisukeYamashita:
### Issue To Be Solved

Currently, the Badger is only supported as an DB and all the data are written to the local disk.
To use this emulator as a development network, in my company, we are planning to deploy this emulator and the dev wallet to the GKE cluster.

But since the Badger makes the workload stateful, we need to create a Statefulset with persistence volume to store the data.
It will make it easier to manage if the emulator can run in stateless manner because Kubernetes workloads are basically designed as stateless.

### (Optional): Suggest A Solution

I saw this article [Supporting digital ownership and decentralization with Dapper Labs and Google Cloud](https://cloud.google.com/blog/topics/startups/scaling-blockchain-and-nfts-with-google-cloud) saying that the Dapper, inc has now partnerships with GCP. 

Therefore, I think the best candidate will be Cloud Firestore, a NoSQL database from GCP.
I also considered Cloud Spanner but it's not KVS and it's quite expensive 💵

sideninja:
I like this feature but I can't promise you a delivery date from our side. Hopefully, we can work on this soon, but you are also welcome to take a stab at it if you have the resources ofc.

KeisukeYamashita:
Hi, I didn't intend to assign someone for this, I wanted to know the communities and the maintainers opinion and opened a ticket. Thank you for your reply!

bluesign:
I think we can close this as now we support redis backend.

PR #221, opened 2022-10-17
Storage layer refactor & example new storage backend impl

bluesign:
## Description

This is a proposed change to the storage layer, for which I drafted a simple version. We can discuss it, but it simplifies the storage layer and allows quickly adding new storage backends. 

I implemented simple Redis server support for it (mostly related to #130 and similar requests )
redis/store.go is the simple implementation example for new backends. 

Also, I want to move snapshot from badger DB to SQLite or something similar ( badger on windows is a bit problematic )

______

For contributor use:


bluesign:
I forgot to mention this will be a breaking change, as the storage format is changing a bit. We can make it backwards compatible and little ugly, or decide to break compatibility.

sideninja:
Sorry for taking longer, but will soon try to review this 🙏

storage/encoding.go
```diff
@@ -71,8 +71,12 @@ func decodeTransactionResult(result *types.StorableTransactionResult, from []byt
 	return cbor.Unmarshal(from, result)
 }
 
-func encodeUint64(v uint64) ([]byte, error) {
-	return em.Marshal(v)
+func encodeUint64(v uint64) []byte {
```
sideninja, line 74: NIT: I feel this should be called `mustEncodeUInt64`

storage/redis/store.go
```diff
@@ -0,0 +1,126 @@
...
+ * limitations under the License.
+ */
+
+package redis
+
+import (
+	"context"
+	"encoding/hex"
+	"fmt"
+	redis "github.com/go-redis/redis/v8"
```
sideninja, line 25: NIT: this alias is not needed

storage/redis/store.go
```diff
@@ -0,0 +1,126 @@
...
+		options: options,
+		rdb:     redis.NewClient(options),
+	}
+	store.DataBackend = store
+	store.KeysBackend = &storage.StorageBackendKeysImpl{}
+
+	return store, nil
+}
+
+func (s *Store) GetBackend() storage.StorageBackendData {
```
sideninja, line 52: I'm not sure if this is supposed to be exported.

storage/store.go
```diff
@@ -65,17 +82,321 @@ type Store interface {
...
+	IdentifierKey(id flowgo.Identifier) []byte
+	EventKey(blockHeight uint64, txIndex, eventIndex uint32, eventType flowgo.EventType) []byte
+}
+
+type StorageBackendData interface {
+	GetBytes(ctx context.Context, store string, key []byte) ([]byte, error)
+	SetBytes(ctx context.Context, store string, key []byte, value []byte) error
+	SetBytesWithVersion(ctx context.Context, store string, key []byte, value []byte, version uint64) error
+	GetBytesAtVersion(ctx context.Context, store string, key []byte, version uint64) ([]byte, error)
+}
```
sideninja, line 113: Very nice!!! I really like this refactor.
One note is to avoid exporting this as I don't believe it's needed outside.

NIT: I feel you could rename them a bit tho: remove storage as it's already part of the package name so you have `storage.StorageBackendKeys` -> `storage.BackendKeys` also maybe `storage.KeyGetter` or `storage.KeyGenerator`, but even if just `BackendKeys`is fine I guess just dropping the storage as it's stuttering https://michaelwhatcott.com/go-code-that-stutters/. The same goes for StorageBackendData it could just be BackendData. It would also make sense if this was broken into two interfaces `DataGetter` and `DataSetter`.

storage/store.go
```diff
@@ -65,17 +82,321 @@ type Store interface {
...
+
+type StorageBackendKeysImpl struct {
```
sideninja, line 115: This `Impl` has very Java vibe. I would rather call it just `Keys` or something like that and then you also don't have to keep repeating keys in method names, so it can be `Keys.Storage()`. Any way you see fit but the `Impl` suffix doesn't feel idiomatic in Go. Also again I don't think this should be exported.

storage/store.go
```diff
@@ -65,17 +82,321 @@ type Store interface {
...
+	return []byte(fmt.Sprintf(
+		"%032d-%032d-%032d-%s",
+		blockHeight,
+		txIndex,
+		eventIndex,
+		eventType,
+	))
+}
+
+type StoreImpl struct {
```
sideninja, line 144: Avoid exporting and rename if possible to drop `Impl`, I suggest maybe `defaultStore` or anything like that

sideninja (approved):
This is an awesome refactor, really like abstraction level you put in!!! awesome

storage/redis/store.go
bluesign: I think it can even be deleted, it is left over somehow I missed.

storage/store.go
bluesign: should I flatten redis , badger stores and move to storage level ?

sideninja: I don't mind them being there, I think it shows nice isolation of each implementation.

storage/store.go
```diff
@@ -39,23 +54,25 @@ import (
 //
 // Implementations must be safe for use by multiple goroutines.
 type Store interface {
+	LatestBlockHeight(ctx context.Context) (uint64, error)
 
 	// LatestBlock returns the block with the highest block height.
-	LatestBlock() (flowgo.Block, error)
+	LatestBlock(ctx context.Context) (flowgo.Block, error)
```
sideninja, line 60: nice! good idea

sideninja (approved):
LGTM!

storage/store.go
bluesign: Yeah I mean in the context of exporting, I was exporting for then to be able to use, but my go knowledge is limited, is there another way?

sideninja: But where do you use it? I didn't find any usage. If you think it needs to be used then leave it exported but I only saw usage of the Store interface.

sideninja:
@bluesign just make sure to resolve the conflict and we merge it.

bluesign:
@sideninja can you assist me here a bit, I am lost in git :(

Issue #219, opened 2022-10-14
[FEATURE] No History


bluesign:
### Issue To Be Solved

In some projects, we don't care about old chain history ( playground, testing etc ) I am thinking a flag ( actually, I have one I implemented as --no-past ) but looking for a nicer name.

This flag will prevent `script execution` and `getAccount` on previous blocks, but it will save a lot of space and memory.

bluesign:
@sideninja, do you have any comment on this? should I make a PR? I think it is also a good update for the playground backend.

bluesign:
I mean the register history, as far as I understand, we have something like this:

- Let's say I have String "deniz" in storage/test ( at block 1 ) 
- at block 2 I delete this, and make it "sideninja" 

this "deniz" stays forever in the database/memory.

sideninja:
I understand. Do you think that's valuable just in terms of space optimization? also, where do you see this playing an important role?

bluesign:
it saves a lot of space, I think mostly CI jobs ( big tests, low memory CI, etc.) Also, it has good synergy with the snapshot feature, we can have tiny snapshots. ( I am planning also sqlite backend support for the emulator as backend changes are merged ) I don't think executeScriptAtBlock is used that much tbh.

bluesign:
I think this will not be too much saving in the short run, and will conflict with rollback. so closing.

PR #222, opened 2022-10-17
Multi tenant emulator 

### LINKED ISSUE #207: [FEATURE] Multi tenant emulator
### Issue To Be Solved

In some cases, multi-tenant support can be valuable to emulator. One emulator instance, running multiple blockchains. 

I opened this issue for to start a discussion, if it is needed like I suspect, I can do the implementation.

bluesign:
Closes #207 

## Description

This is a small PR allowing using HTTP header `Emulator` to quickly start a new blockchain instance and switch between them with `Emulator: [someIdentifier]`

This is also on the idea phase, and we can discuss over it, but I think it can be very useful, as flow-js-testing etc, always launching new emulator instances, starting and stopping emulators. 

This can also let few users to use a single emulator server concurrently. (no port clashes etc) 

______

For contributor use:

### DIFF
--- a/server/admin.go
+++ b/server/admin.go
@@ -115,6 +115,10 @@ func (h *HTTPServer) Stop() {
 	_ = h.httpServer.Shutdown(context.Background())
 }
 
+func (h *HTTPServer) Server() *http.Server {
+	return h.httpServer
+}
+
 func wrappedHandler(wrappedServer *grpcweb.WrappedGrpcServer, headers []HTTPHeader) http.HandlerFunc {
 	return func(res http.ResponseWriter, req *http.Request) {
 		setResponseHeaders(&res, headers)
--- a/server/backend/backend.go
+++ b/server/backend/backend.go
@@ -54,10 +54,12 @@ func (b *Backend) SetEmulator(emulator Emulator) {
 	b.emulator = emulator
 }
 
+// Allows backend to create emulator on demand
 func (b *Backend) SetEmulatorCreator(creator func(name string) (*emulator.Blockchain, error)) {
 	b.createEmulator = creator
 }
 
+// Switches active blockchain for current request
 func (b *Backend) SwitchEmulator(name string) error {
 	emulator, ok := b.emulators[name]
 	if ok {
--- a/server/backend/backend.go
+++ b/server/backend/backend.go
@@ -42,9 +42,11 @@ import (
 // Backend wraps an emulated blockchain and implements the RPC handlers
 // required by the Access API.
 type Backend struct {
-	logger   *logrus.Logger
-	emulator Emulator
-	automine bool
+	logger         *logrus.Logger
+	emulator       Emulator
+	emulators      map[string]Emulator
+	automine       bool
+	createEmulator func(string) (*emulator.Blockchain, error)
 }
 
 // SetEmulator hotswaps emulator for state management.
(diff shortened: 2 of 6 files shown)


turbolent:
What are the advantages of this built-in support for multiple emulators compared to starting multiple emulators on different ports?

bluesign:
> What are the advantages of this built-in support for multiple emulators compared to starting multiple emulators on different ports?

@turbolent 

2 big advantages:

- It is cheaper than launching emulator; especially when resource-constrained conditions ( like CI running tests or wasm emulator on the web ) 

- No need for heavy emulator management. For example, in flow-js-testing: There is a need for big management code they have to write: They need to find empty ports, launch the emulator with them and switch the FCL config to that port ( but as the config is singleton there, they need to run tests inband, not parallel, etc.) 

Actually, this idea came from trying to speed up flow-js-testing. ( and run tests in parallel easily )

janezpodhostnik:
I see the utility of this when running in a test setting. 

Not sure if using headers to switch versions would be my first choice. Maybe it would be easier to just bind multiple ports each to their own emulator (own blockchain, not own emulator process).

With the current code I'm also concerned about concurrent access. Can we add some tests?

bluesign:
@janezpodhostnik thank you very much for the comments, this is to kick off the discussion PR (issue didn't get much attention I wanted to show what I mean more clearly ),  I was gonna make this draft PR, but somehow I am GH cli noob. I will add tests and concurrency improvements for sure. 

For the headers part: This is the standard API key model, but I just recognized that I didn't give enough context within the issue and PR about the direction I am targeting. So let me briefly explain what I have in my mind. 

The emulator is the core component when doing the development. There are a few pain points I recognized where users always struggle when developing and testing dapps, such as: 

- slow testing ( especially with js-testing ) 
- state management ( we have some state management, but I am not proud of it ) 
- different cadence versions ( especially testing with upcoming cadence versions on breaking changes ) 

For the first target ( slow testing ), multiple ports are a solution. Eventually, it is a bit tricky and not sustainable in scale ( think of the playground using a backend for many users/projects). In that direction, with new storage backend improvements https://github.com/onflow/flow-emulator/pull/221 I am trying to find a suitable backend that can scale and with https://github.com/onflow/flow-emulator/issues/219 trying to make it cost-viable. ( I have a few playground alike projects like jupyter notebook etc.) 

For the state management, this provides at least context switching for free, forking part has to be worked on for sure. 

For the different cadence versions, it is a bit of another direction ( or a long-term goal ); with these changes, it leads to the possibility of running an emulator as a SaaS ( think of it as a disposable emulator server where you can connect to an instance of the emulator by pointing out your client with an API key + project identifier ) and run your tests. So basically, just changing the endpoint, you can run it all on a different instance of the emulator running with the upcoming version of cadence.

sideninja:
@bluesign it feels to me like this could/should be handled on another layer, not by the emulator itself. I'm afraid it adds complexity in the codebase that is very specific to running tests maybe. I understand the pain point is bootstrapping the emulator so wouldn't a solution potentially be having a pool of bootstrapped emulator instances and then doing some kind of load balancing between them where there can be a policy on the load balance to know which instance it forwards to. 
I'm a bit on the fence with this one tbh. So can we try and look into any solutions on another layer if we can avoid adding the complexity of this here, especially since I'm not too convinced this is a common problem yet.

bluesign:
I understand the concerns, and to be honest, I don't have too hard an opinion on this feature too. We can maybe drop this PR. My only concern was every consumer ( js-testing, overflow, playground ( my version ) ) has to manage same thing themselves.

sideninja:
> I understand the concerns, and to be honest, I don't have too hard an opinion on this feature too. We can maybe drop this PR. My only concern was every consumer ( js-testing, overflow, playground ( my version ) ) has to manage same thing themselves.

But is that the current issue with overflow and playground? For playground I can say that we use emulator as a module and have instances bootstrapped as part of the instance pool so we can always have them ready: https://github.com/onflow/flow-playground-api/blob/master/blockchain/pool.go#L28
I will move this to draft for now ok? maybe we can revisit if more usage will require this but for now I'm still not convinced it justifies the added complexity and possible issues with maybe some concurrency.

PR #224, opened 2022-10-18
Add interaction logs

### LINKED ISSUE #223: [Feature] - Add logs from transaction and scripts
### Issue To Be Solved
Since logs calls are not preserved on blockchain they are excluded from Access API, but they are crucial for debugging, testing and development process.

### Suggest A Solution
- Call for transaction result shall have `Logs` field
- Script execution result shall have `Logs` field 

### Context
Testing library is trying to extract and parse log statement from `stdout` stream, which is subpar and prone to errors.

MaxStalker:
## Description
- Transaction result can't be returned before new block is commited and all changes done by transaction are calculated and stored. I've added new admin endpoint, which will allow to access logs by transaction id at `/emulator/logs/{txId}`. This will keep Backend in complience with access.API interface and (technically 😅) shouldn't break existing solutions 🤞 
- Script result can be returned with HTTP request, so we create new type, which holds all the fields of ScriptResult, add new field `Logs` and then encode it as json

## Drawbacks
- gRPC endpoints are not implemented 😬

@sideninja need your advice on how to properly implement/improve this feature 🙇 

Closes #223 
______

For contributor use:

### DIFF
--- a/server/backend/backend.go
+++ b/server/backend/backend.go
@@ -592,6 +592,11 @@ func (b *Backend) CommitBlock() {
 	}).Debugf("📦  Block #%d committed", block.Header.Height)
 }
 
+type ScriptResultWithLogs struct {
+	types.ScriptResult
+	Logs []string
+}
+
 // executeScriptAtBlock is a helper for executing a script at a specific block
 func (b *Backend) executeScriptAtBlock(script []byte, arguments [][]byte, blockHeight uint64) ([]byte, error) {
 	result, err := b.emulator.ExecuteScriptAtBlock(script, arguments, blockHeight)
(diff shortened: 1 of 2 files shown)


bluesign:
It seems it should work with grpc requests too. 

Although I think returning script logs with emulator API endpoint like transaction can be good, it makes it super complicated with scriptID collision ( and parallel tests ) 

Also with `ScriptResultWithLogs`, we are returning totally different result it seems.

sideninja:
@MaxStalker can you check what is going on with tests? 
On the first glance I think this might be problematic with concurrent execution. But I see the need for this functionality. Will try to share some thoughts soon.

bjartek:
fair enough

sideninja:
> Transaction result can't be returned before new block is commited and all changes done by transaction are calculated and stored.

Can you elaborate why this is problematic?

sideninja:
Just converted to draft until tests are passing and it's ready to review.

bluesign:
I added support for tx logs via API, scripts logs still hanging but this PR is very stale, closing, @MaxStalker if you want me to reopen it please just say.

PR #236, opened 2022-11-09
Expose option for contract removal option

sideninja:
## Description
Running an emulator programmatically it restricts removing of contracts, probably because that value is not set as part of the service contracts. Exposing to explicitly set that helps.
______

For contributor use:

### DIFF
--- a/blockchain.go
+++ b/blockchain.go
@@ -138,6 +138,7 @@ type config struct {
 	TransactionExpiry            uint
 	StorageLimitEnabled          bool
 	TransactionFeesEnabled       bool
+	ContractRemovalRestricted    bool
 	MinimumStorageReservation    cadence.UFix64
 	StorageMBPerFLOW             cadence.UFix64
 	Logger                       zerolog.Logger
@@ -338,6 +339,15 @@ func WithTransactionFeesEnabled(enabled bool) Option {
 	}
 }
 
+// WithContractRemovalRestricted restricts/allows removal of already deployed contracts.
+//
+// The default is provided by on-chain value.
+func WithContractRemovalRestricted(restricted bool) Option {
+	return func(c *config) {
+		c.ContractRemovalRestricted = restricted
+	}
+}
+
 // WithTransactionValidationEnabled enables/disables transaction validation.
 //
 // If set to false, the emulator will not verify transaction signatures or validate sequence numbers.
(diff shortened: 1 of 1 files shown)


turbolent:
Contract removal was disabled on all networks with the release of Secure Cadence, as removing a contract and then re-adding a contract with the same name allows circumventing the contract updatability checks, which ensure that a contract update is safe, in the sense that it is supported by the interpreter. 

For example, adding a new field is not supported, as existing stored data does not have the field and there is no migration functionality available (yet) to backfill existing fields. 

There's currently no immediate plan to solve contract removal, and discussions in the Cadence Language Design meetings regarding this current limitation argued that we might want to completely remove the contract removal functionality.

Are we sure we want to expose this option? Developers will likely enable the option to allow removal, use it incorrectly, and run into crashers locally. Even if they use it in a way that does not crash the environment, they will likely be surprised the functionality won't be available on any network (incl. Testnet).

turbolent (approved):

bluesign:
@sideninja @turbolent  I think I agree with both of you in this; 

- I think the emulator should have super powers
- Also it shouldn't do anythinthat cannot be done on mainnet/testnet 

I propose `import emulator` ( emulator stdlib) and putting everything not in mainnet/testnet in that lib.  So we can use things like `emulator.remove_contract`, also we can expose a lot of useful stuff. Also when users are using explicitly writing `emulator.something` gives the impression that it will work only on the emulator, so it will not generate any complaints.

turbolent:
@bluesign Great idea! Only small issue with this is the additional effort required to implement this

bluesign:
> @bluesign Great idea! Only small issue with this is the additional effort required to implement this.

If we can manage the stdlib expose from host to cadence, the rest should be pretty easy. But actually, as an added benefit this can clean up a lot of stuff from Cadence ( move them to fvm, emulator etc ) 

But yeah eventually in the end, additional effort is needed for sure.

sideninja:
> > @bluesign Great idea! Only small issue with this is the additional effort required to implement this.
> 
> If we can manage the stdlib expose from host to cadence, the rest should be pretty easy. But actually, as an added benefit this can clean up a lot of stuff from Cadence ( move them to fvm, emulator etc )
> 
> But yeah eventually in the end, additional effort is needed for sure.

Should we open this as an issue (improvement) since I don't think it's feasible addressing with this PR.

PR #237, opened 2022-11-09
Separate listen and start methods on the emulator server

sideninja:
## Description
If the emulator is used as a package programmatically it's impossible to know when the server is actually up and listening since the Start method is blocking and doesn't signal when the server is ready. This PR adds the Listen method so you can explicitly call it and from that point onwards treat the server as ready. 

______

For contributor use:

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -43,11 +43,12 @@ type EmulatorServer struct {
 	config     *Config
 	backend    *backend.Backend
 	group      *graceland.Group
+	listeners  []listener
 	liveness   graceland.Routine
 	storage    graceland.Routine
-	grpc       graceland.Routine
-	admin      graceland.Routine
-	rest       graceland.Routine
+	grpc       *GRPCServer
+	rest       *RestServer
+	admin      *HTTPServer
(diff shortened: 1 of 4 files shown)


server/server.go
```diff
@@ -192,7 +197,29 @@ func NewEmulatorServer(logger *logrus.Logger, conf *Config) *EmulatorServer {
...
 
+// Listen starts listening for incoming connections.
+//
+// After this non-blocking function executes we can treat the
+// emulator server as ready.
+func (s *EmulatorServer) Listen() error {
+	s.listeners = make([]listener, 0)
+	s.listeners = append(s.listeners, s.grpc)
+	s.listeners = append(s.listeners, s.rest)
+	s.listeners = append(s.listeners, s.admin)
```
turbolent, line 208: Maybe shorten this a bit:

```suggestion
	s.listeners = []listener{s.grpc, s.rest, s.admin}
```

turbolent (approved):
Nice!

server/server.go
sideninja: lol

PR #238, opened 2022-11-09
Revert "Backend cleanup ( adapter removal )"

sideninja:
Reverts onflow/flow-emulator#220

Updating this into CLI made me realize that we probably shouldn't change the types of the backend. It's ok to remove the adapter as that's not used anywhere but changing types to flow-go types breaks any integration of the emulator. Also, I think returning SDK types makes it easier to use.

cc @bluesign

### DIFF
--- a/server/backend/adapter.go
+++ b/server/backend/adapter.go
@@ -1,0 +1,220 @@
+/*
+ * Flow Emulator
+ *
+ * Copyright 2019 Dapper Labs, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


turbolent:
What is the reason for the revert?

turbolent (approved):
Please add the reasoning to the PR description, so we can keep track of why it was (temporarily) reverted. Thanks!

sideninja:
@bluesign what do we do here now? I think returning flow-go native types from emulator would be problematic for consumers. Can we have some actions here so things don't get stale 🙏

bluesign:
I think the naming of the backend confuses me a bit, but my understanding is the best option:

- move emulator related functionality to emulator instead of backend ( there is no difference there ) 
- emulator with go-sdk types can service flowkit and other consumers 
- we can rename the backend to something more meaningful, and use flow-go types there. ( as rest API and grpc API directly using that ) maybe backend can be 'adapter' ( adapting emulator to grpc and rest API )

sideninja:
I like this plan @bluesign. I will extract it into an issue and revert the change by the previous PR since it would break integration. Is that good with you?

PR #241, opened 2022-11-17
Shapshots list admin endpoint 

### LINKED ISSUE #228: Split snapshot management into multiple HTTP endpoints
### Issue To Be Solved

Currently, there is a single endpoint for managing emulator snapshots: 
https://github.com/onflow/flow-emulator/blob/937995270076dd8d48a25df39d8f509c89bfec93/server/emulator.go#L90-L137

This endpoint performs multiple actions:
1. Creates a new snapshot
2. Jumps to an existing snapshot
3. Retrieves an existing snapshot

I think this is not ideal because:
1. It's confusing from the API consumer standpoint
4. Adds unnecessary API usage constraints, like:
   - API consumers can't check if the snapshot exists without risking that this will also trigger the "move state to snapshot" action
   - ...?

### Suggest A Solution

I think it would make sense to split those actions into separate REST endpoints:

1. `POST /emulator/snapshots/{name}` - create (or reverts to existing) snapshot
2. `PUT /emulator/snapshots/{name}` - revert to an existing snapshot (optional - could also work to just combine 1. and 2. into a single endpoint)
3. `GET /emulator/snapshots/{name}` - retrieve existing snapshot

### Context

This improvement would allow better snapshot management in [Flowser desktop app](https://github.com/onflowser/flowser), as it would allow us to:
- better handle snapshot errors and help discover unexpected bugs (an example of such bug is https://github.com/onflow/flow-emulator/issues/227)
- check whether the snapshots that were created (which are also stored/cached in our database) still exist at different points in time (e.g. user could manually remove `flowdb` folder, introducing a state conflict)

bluesign:
Closes #228 

## Description

I was thinking 2 endpoints:

GET /snapshots to list all snapshots, in any case we should add, I guess, it seems pretty nice to have.
GET /snapshots/{snapshotID} Retrieve existing one is also not a problem. ( also 404 on fail )

But it will not be backwards compatible. But if we decide like that ( and move create to POST verb ) I can update the PR

______

For contributor use:

- [X] Updated relevant documentation


sideninja:
I would say that if we are doing changes in API I would make the API to be more RESTful:
`GET /snapshots` list all
`POST /snapshots` create a new snapshot, and provide a name in body
`PUT /snapshots/{id}` restore a specific snapshot

sideninja:
@bluesign is this complete and ready for review? sorry I guess we missed it.

bluesign:
Yeah should be ready, Sorry I forgot this too

server/emulator.go
```diff
@@ -84,49 +89,119 @@ func (m EmulatorAPIServer) CommitBlock(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+}
+func (m EmulatorAPIServer) SnapshotList(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	badgerStore := (*m.storage).Store().(*badger.Store)
```
sideninja, line 95: will this work with multiple store implementations? should you do a check whether the type is indeed badger.

server/emulator.go
```diff
@@ -84,49 +89,119 @@ func (m EmulatorAPIServer) CommitBlock(w http.ResponseWriter, r *http.Request) {
...
+		Context: name,
+	}
+
+	bytes, err := json.Marshal(blockResponse)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
```
sideninja, line 139: content type to json?

server/emulator.go
```diff
@@ -84,49 +89,119 @@ func (m EmulatorAPIServer) CommitBlock(w http.ResponseWriter, r *http.Request) {
...
+func (m EmulatorAPIServer) SnapshotCreate(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	name := r.FormValue("name")
+
+	if name == "" {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	switch (*m.storage).Store().(type) {
```
sideninja, line 186: nit: a if check would be nicer to just return fast, something like:
```go
badger, isBadger := s.(string)
if !isBadger {
    return "error"
}
```

server/emulator.go
```diff
@@ -84,49 +89,119 @@ func (m EmulatorAPIServer) CommitBlock(w http.ResponseWriter, r *http.Request) {
...
-			Height:  int(block.Header.Height),
-			BlockId: block.Header.ID().String(),
-			Context: name,
+		if slices.Contains(contexts, name) {
+			w.WriteHeader(http.StatusConflict)
+			return
 		}
-
-		bytes, err := json.Marshal(blockResponse)
+		err = badgerStore.JumpToContext(name, true)
```
sideninja, line 198: I don't completely understand if this is necessary to jump? aren't we already at the point of snapshot?

sideninja, line 198: This might be NIT but is it ok that creation and jumping is all in one step? why I mean is because if we just want to create we already are at that point so doing an actual jump as well feels too much. I think this might have been because before you had both operation as one (either create or jump) but now that we divided those operations in API this might also better be divided.

server/emulator.go
bluesign: Actually my plan is to migrate snapshots to sqlite, they are small, and self contained. 

Badger has windows big file size problem (or windows has the problem according to badger guys)

Some portable snapshot format (still we can use git, but also can provide a download api ) can be pretty useful.

server/emulator.go
bluesign: This is little tricky, file analogy is: 'save as' , git analogy is 'branch & switch & commit' 

All OS file actions, switch after save as, git is also require you to switch before commit. 

But we can change the behavior, I don't have strong opinion on this UX.

bluesign:
@sideninja this one also I will handle Monday if not urgent

server/emulator.go
bluesign: do we need that?

server/emulator.go
sideninja: I see. Yeah I will leave it to your judgment. Not a big thing.

sideninja (approved):

sideninja:
@bluesign are we ok to merge?

bluesign:
> @bluesign are we ok to merge?

sorry I missed this, this can be merged.

bluesign:
@sideninja very good idea, I will make a PR

PR #251, opened 2022-12-08
Added dependencies.yml action

gpavlov2016:
## Description

Added a github action to build flow-cli from main with flow-emulator dependency replaced by sources in this commit. This step makes sure that the current update to emulator doesn't break the flow-cli
______

For contributor use:

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -1,0 +1,82 @@
+name: Ensure downstream dependencies are not broken by this commit
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branches: [ gpavlov/dependencies ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  flow-cli-main:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Save current timestamp to env var
+        run: |
+            echo "TIMESTAMP=$(date)" >> $GITHUB_ENV
+      
+      - name: Get Emulator sources
+        uses: actions/checkout@v3
+        with:
+          repository: onflow/flow-emulator
+          path: flow-emulator
+          ref: ${{ github.ref }}  
+
+      - name: Get flow cli sources
+        uses: actions/checkout@v3
+        with:
+          repository: onflow/flow-cli
+          path: flow-cli
+
+      ###
+
+      - uses: actions/setup-go@v1
+        with:
+          go-version: '1.19'
+      
+      - name: Install build dependencies 
+        run: |
+          sudo apt-get update -y 
+          sudo apt-get install build-essential zip cmake -y 
+
+      ###
+
+      - name: Replace imports with local sources in flow-cli project
+        run: | 
+          cd flow-cli
+          go mod edit -replace github.com/onflow/flow-emulator=../flow-emulator
+
+      - name: Build emulator
+        run: |
+          cd flow-emulator
+          go mod tidy
+          go build ./cmd/emulator
+
+      - name: Build flow-cli
+        run: |
+          cd flow-cli
+          go mod tidy
+          make versioned-binaries
+
+      - name: Copy flow binary
+        run: cp flow-cli/cmd/flow/flow-x86_64-linux- /usr/local/bin/flow && ls /usr/local/bin/flow
+
+      - name: Get Flow Version
+        run: |
+          echo Flow version:
+          flow version
+
+      - name: Upload flow-cli artifacts
+        # if: ${{ env.ACTIONS_RUNTIME_TOKEN }}
+        uses: actions/upload-artifact@v3
+        with:
+          name: flow-cli
+          path: flow-cli/cmd/flow
+  
+      - name: Upload emulator artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: flow-emulator
+          path: flow-emulator/emulator
+


.github/workflows/dependencies.yml
```diff
@@ -0,0 +1,82 @@
+name: Ensure downstream dependencies are not broken by this commit
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branches: [ gpavlov/dependencies ]
```
justinbarry, line 7: Did you mean to hardcore this in here?

justinbarry (approved):
One comment. Otherwise approved.

bluesign:
Is this the correct approach? Shouldn't the emulator sometimes break flow-cli? 

I feel like, it will end up with a 'die hard' situation[0] as currently we have with flow-go and flow-emulator. 

- update emulator ( oh it breaks cli cant merge ) 
- go to cli, update cli dependency to commit hash of this emulator PR, fix changes 
- go to emulator, make a release
- go to cli, update cli dependency to release 

[0] https://www.youtube.com/watch?v=BVtQNK_ZUJg

.github/workflows/dependencies.yml
gpavlov2016: I am expecting few more updates to the script (for example notification to slack) so I am keeping it here to enable quick experementation

gpavlov2016:
> Is this the correct approach? Shouldn't the emulator sometimes break flow-cli?
> 
> I feel like, it will end up with a 'die hard' situation[0] as currently we have with flow-go and flow-emulator.
> 
> * update emulator ( oh it breaks cli cant merge )
> * go to cli, update cli dependency to commit hash of this emulator PR, fix changes
> * go to emulator, make a release
> * go to cli, update cli dependency to release
> 
> [0] https://www.youtube.com/watch?v=BVtQNK_ZUJg

If we have a situation where the emulator has to break cli for a long time it means that we in fact managing two active versions of both. In this case the solution is to open a new branch on both repos and manage them in parallel testing the dependency chains separately in each branch.

.github/workflows/dependencies.yml
sideninja: I suggest keeping it as draft until it's ready to review.

sideninja (commented):
I might misunderstand this PR's intent, but if I understand correctly, you are trying to check when a new PR is made on emulator to build CLI and other tools to see if it builds with the changes in this PR. If that is the case then I don't believe this will work. The problem is that anytime there is gonna be an API change it will fail, at the same time there's no way around making this succeed when there is an API change since it requires you adapt the CLI beforehand but at the same time you are pulling from master but the adaptation will never be on master since this PR is not yet accepted. You created a messy circural dependency with this. I might misunderstand tho, so please explain.

gpavlov2016:
> 

The main goal is to identify PRs that break CLI. What we do with this is a matter of policy and should be adjusted based on the frequency and availability. At this stage there is not intent to block PRs due to failure of this test.
In the next step I will add slack notification when the test fails and will be monitoring it with the goal to coordinate CLI update that will fix the build. 
I think we all can agree on two things:
- The sooner we address the breaking change on the CLI side, the less effort it will require
- It is better to know that the build fails rather than being surprised by it when we need to release the next CLI version

bluesign:
I am Github noob but I think solution for your problem is Dependabot and semver @gpavlov2016. PR hook is a really bad idea for what you are trying to do. 

You should define `Dependabot` action (or some similar action) on CLI repo, which will watch `Releases` ( you can set interval ) from `flow-emulator`. Then alert you if something breaks. 

> The sooner we address the breaking change on the CLI side, the less effort it will require
Yeah totally agreed

> It is better to know that the build fails rather than being surprised by it when we need to release the next CLI version
I think best is to auto create an issue on flow-cli repository ( with Dependabot on similar )  on new `Emulator` release.

gpavlov2016:
> I am Github noob but I think solution for your problem is Dependabot and semver @gpavlov2016. PR hook is a really bad idea for what you are trying to do.
> 
> You should define `Dependabot` action (or some similar action) on CLI repo, which will watch `Releases` ( you can set interval ) from `flow-emulator`. Then alert you if something breaks.
> 
> > The sooner we address the breaking change on the CLI side, the less effort it will require
> 
> Yeah totally agreed
> 
> > It is better to know that the build fails rather than being surprised by it when we need to release the next CLI version
> 
> I think best is to auto create an issue on flow-cli repository ( with Dependabot on similar ) on new `Emulator` release.

Unfortunately, detecting breaking changes when emulator is being released is too late. By that time the build might have been broken for several weeks, and the person that made the changes is unavailable to help. Moreover, usually when emulator is released, the CLI needs to be released as well which will not be possible due to breaking changes.

I can't find any significant problem with adding additional non-blocking check to the CI, would you mind elaborating why do you think it's a bad idea?

sideninja:
> Unfortunately, detecting breaking changes when emulator is being released is too late. 

But if the change isn't yet released how would we make the PR on the dependant tool? I think dependabot actually waits for a release because at that point you can act and that's intentionally by design it's not because they didn't think of this. Doing that sooner would require targeting a specific commit has as a version in that branch which is not a good practice. Or we will be forced to have PRs that will sit there for several weeks (like you said it can take) and might get outdated during that time as changes will be done on other parts. I think if merged PRs sit there for weeks then the core problem is we should be doing more smaller releases, which we can also automate.

> rather than being surprised by it when we need to release the next CLI version

We are not surprised by the new version of the emulator breaking the CLI during release. That is a natural process when updating a dependency, you make a PR to update the version of the emulator and do any changes that potentially broke APIs. The only "surprises" that came were platform-specific issues that were only detected on that platform.  

> I can't find any significant problem with adding additional non-blocking check to the CI, would you mind elaborating why do you think it's a bad idea?

I think the issue is it will force us to ignore failed CI actions.

PR #257, opened 2022-12-22
Contract Removal Flag

sideninja:
## Description
Allow contract removals when using the `--contract-removal` flag on the emulator. Emulator should act in two ways, first is to precisely match the behavior of the mainnet for testing, but during the initial development phase, it has to allow different development features that are not possible on mainnet in order to make the development process easy. Removing contract is for sure one of them, which would allow devs to update contracts as they develop them. 
______

For contributor use:


bluesign:
I think we can make a `--strict` flag to emulate mainnet behavior strictly, but default settings should focus on `ease of development` in my opinion.

sideninja:
> I think we can make a `--strict` flag to emulate mainnet behavior strictly, but default settings should focus on `ease of development` in my opinion. ( we don't have transaction fees enabled default for example )

I was thinking of something very similar. I don't want to make it part of this PR tho. But strict flags sounds the way to go to me. And that should be explicitly required.

janezpodhostnik (approved):

PR #258, opened 2022-12-22
Remove logs from debug

sideninja:
This is super subjective so I will include more people to share opinions and have no trouble closing this PR, but I've noticed that when I develop something I use logs during development to see how things work and while using an emulator I always find myself forgetting to use the verbose mode and even when I do there's a lot of extra stuff that I don't want to see. I feel if a developer is using Logs they are doing so explicitly and then that not being part of the default output of the emulator might be a confusing experience. Imagine using a Println and not being printed if not used with verbose mode. 

cc @janezpodhostnik @turbolent @bluesign @bjartek 
______

For contributor use:


bluesign:
I agree with this change and also suggest to have better Logs. ( for example when transaction is failed Logs are gone ) But actually failure cases are the usually the ones I need logs.

sideninja:
> I agree with this change and also suggest to have better Logs. ( for example when transaction is failed Logs are gone ) But actually failure cases are the usually the ones I need logs.

I feel this deserves a separate issue. But I agree 100%.

turbolent (approved):
Agreed, debug is not the right level, its purpose is for debugging the emulator, not a contract running in the emulator 👍

PR #263, opened 2023-01-04
update emulator test

pattyshack:
Closes #???

## Description

fvm returns additional errors after upgrading, due to preprocessing.

    
    
    ```
    [Error Code: 1101] error caused by: 2 errors occurred:
    * transaction preprocess failed: [Error Code: 1101] cadence runtime error: Execution failed:
    error: authorizer count mismatch for transaction: expected 0, got 1
    --> 1ae04fd007e8137d3a433f2248b34770a249f148fcd2405480b2c0076d5129d8
    
    * transaction verification failed: checking frozen account failed: cannot check account freeze status: [Error Code: 1201] account not found for address 01cf0e2f2f715450
    ```
______

For contributor use:

### DIFF
--- a/transaction_test.go
+++ b/transaction_test.go
@@ -716,7 +716,7 @@ func TestSubmitTransaction_EnvelopeSignature(t *testing.T) {
 
 		script := []byte(`
 		  transaction {
-		    prepare() {}
+		    prepare(signer: AuthAccount) {}
 		  }
 		`)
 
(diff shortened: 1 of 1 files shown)


transaction_test.go
```diff
@@ -742,7 +742,7 @@ func TestSubmitTransaction_EnvelopeSignature(t *testing.T) {
 		result, err := b.ExecuteNextTransaction()
 		assert.NoError(t, err)
 
-		assert.True(t, fvmerrors.IsAccountNotFoundError(result.Error))
+		assert.ErrorContains(t, result.Error, "account not found")
```
janezpodhostnik, line 745: does it make sense to check that the err code is present instead?

janezpodhostnik (approved):

transaction_test.go
pattyshack: IsAccountNotFoundError internally calls HasErrorCode, which right now only checks the error code in the first error (1101) and ignore the error codes in the other errors when fvm returns multiple errors.  Hence the failure.  we can probably update HasErrorCode to check for secondary error codes as well

transaction_test.go
```diff
@@ -716,7 +716,7 @@ func TestSubmitTransaction_EnvelopeSignature(t *testing.T) {
 
 		script := []byte(`
 		  transaction {
-		    prepare(signer: AuthAccount) {}
+		    prepare() {}
```
SupunS, line 719: I think these two should be two tests: (1) signing with an invalid account. (2) signing when the transaction does not expect a signer-parameter.
Maybe keep this test as is, and add a new test for (2)?

pattyshack: sure.  i can do that

pattyshack:
I've split the test into two cases as suggested.  I'll merge this tomorrow assuming no objections.

SupunS (approved):
Thank you!

turbolent (approved):

PR #266, opened 2023-01-11
Update to current flow-go master and `m4ksio/batch-removal`

m4ksio:
There are changes in flow-go storage interfaces, so a little change in is require to make emulator compile. 
This allows https://github.com/onflow/flow-go/pull/2535 to pass CI and be merged. 
Once this happen we can change this PR to depend on `flow-go` / `master`


bluesign:
@m4ksio  Can we get rid of headers.go by changing `ByHeightFrom` in blocks.go ?

```go 
func (b *blocks) ByHeightFrom(height uint64, header *flowgo.Header) (*flowgo.Header, error) {
	block, err := b.blockchain.storage.BlockByHeight(context.Background(), height)
	if err != nil {
		return nil, err
	}
	return block.Header, nil
}
```

I don't know much about flow-go, but this seems used for exposing blocks to cadence, so in emulator case, we should be ok with using from storage. 

Also a bit nit but, is there a specific reason not passing tracer in `EnvironmentParams` ?

sideninja (approved):

m4ksio:
> @m4ksio Can we get rid of headers.go by changing `ByHeightFrom` in blocks.go ?
> 
> ```go
> func (b *blocks) ByHeightFrom(height uint64, header *flowgo.Header) (*flowgo.Header, error) {
> 	block, err := b.blockchain.storage.BlockByHeight(context.Background(), height)
> 	if err != nil {
> 		return nil, err
> 	}
> 	return block.Header, nil
> }
> ```
> 
> I don't know much about flow-go, but this seems used for exposing blocks to cadence, so in emulator case, we should be ok with using from storage.

I don't think I have enough knowledge about emulator to answer that with authority. We should probably have this as a separate issues though.

> 
> Also a bit nit but, is there a specific reason not passing tracer in `EnvironmentParams` ?

I don't see any usage of tracing in emulator, using real span would be useless without properly initializing and utilizing entire tracing mechanism, which is far out of scope of this PR.

Issue #274, opened 2023-01-18
Remove headers.go


bluesign:
### Issue To Be Solved

This came up on recent PR #266 

Do we really need headers in emulator? @janezpodhostnik @turbolent 

I think it is only used for cadence `getBlock` functions, as we don't fork chain in emulator, returning block from storage seems to suffice. But I maybe missing something

in blocks.go maybe we can do something like:

```
func (b *blocks) ByHeightFrom(height uint64, header *flowgo.Header) (*flowgo.Header, error) {
       // bluesign: maybe some height check here needed 

       block, err := b.blockchain.storage.BlockByHeight(context.Background(), height)
	if err != nil {
		return nil, err
	}
	return block.Header, nil
}
```

I think this will make emulator - flow-go coupling looser.

AlexHentschel:
Your proposal seems ok. Yet, I don't know when the emulator finalizes blocks. Does it do that immediately when the block is produced?

janezpodhostnik:
> I think it is only used for cadence getBlock functions

That might be the case. In which case what you are proposing makes total sense. I'll take some time to look this over tomorrow.

bluesign:
thanks a lot both of you

janezpodhostnik:
Looks like the use of `headers`  in `headers.go` is in the function you mentioned:

```
func (b *blocks) ByHeightFrom(height uint64, header *flowgo.Header) (*flowgo.Header, error) {
	return environment.NewBlockFinder(b.headers).
		ByHeightFrom(height, header)
}
```

Removing them and rewriting this function would get rid of a lot of code. Getting the block by height, is totally acceptable.

bluesign:
thank you very much @janezpodhostnik , I will make a PR tomorrow.

PR #275, opened 2023-01-18
extension of Headers interface in `flow-go` repo

AlexHentschel:
in PR https://github.com/onflow/flow-go/pull/3824, we are extending the `Headers` interface. Also adding an implementation of the new method `BlockIDByHeight` `headers` implementation in the emulator repo.

### DIFF
--- a/headers.go
+++ b/headers.go
@@ -68,6 +68,15 @@ func (h headers) ByHeight(height uint64) (*flowgo.Header, error) {
 	return block.Header, nil
 }
 
+// BlockIDByHeight returns the block ID that is finalized at the given height.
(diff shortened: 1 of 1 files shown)


bluesign:
@AlexHentschel any chance you can have a look at #274 ? Maybe we can get rid of it in emulator totally

bluesign (approved):

AlexHentschel:
@bluesign what you are proposing seems very reasonable to me. But I don't know the emulator repo at all (this is my first interaction with the code base) ... so probably not the best person to make that call.

janezpodhostnik (approved):

PR #280, opened 2023-01-20
Retire headers.go

### LINKED ISSUE #274: Remove headers.go

bluesign:
Closes #274 

Discussion is in the referenced issue. 

______

For contributor use:


blocks.go
```diff
@@ -68,6 +66,13 @@ func (b *blocks) FinalizedHeader() (*flowgo.Header, error) {
 }
 
 func (b *blocks) ByHeightFrom(height uint64, header *flowgo.Header) (*flowgo.Header, error) {
-	return environment.NewBlockFinder(b.headers).
-		ByHeightFrom(height, header)
+	if height > header.Height {
+		return nil, storage.ErrNotFound
+	}
```
janezpodhostnik, line 71: I think you can also add

```go
if header.Height == height {
    return header, nil
}
```

sideninja (approved):

janezpodhostnik (commented):
Looks good!

Maybe just add a comment mentioning that we don't have to do anything complex, because the emulator doesn't have forks.

blocks.go
bluesign: I think this one is in flow-go , but I will verify

janezpodhostnik: ah, didn't see that. Good catch

janezpodhostnik (approved):

bluesign:
I messed up the branch when rebasing I guess, fixing

janezpodhostnik:
I think you might need to rebase it to the top of master.

bluesign:
@janezpodhostnik I think somehow GitHub showing wrong diff. ( I rebased once again )

bluesign:
ok seems I managed to fix the problem somehow, I think can be merged.

PR #306, opened 2023-01-27
Update fvm api calls

pattyshack:
Closes #???

## Description

______

For contributor use:

### DIFF
--- a/blockchain.go
+++ b/blockchain.go
@@ -30,9 +30,9 @@ import (
 	"github.com/onflow/flow-go/engine/execution/state/delta"
 	"github.com/onflow/flow-go/fvm"
 	fvmcrypto "github.com/onflow/flow-go/fvm/crypto"
+	"github.com/onflow/flow-go/fvm/derived"
 	"github.com/onflow/flow-go/fvm/environment"
 	fvmerrors "github.com/onflow/flow-go/fvm/errors"
-	"github.com/onflow/flow-go/fvm/programs"
 	"github.com/onflow/flow-go/fvm/state"
 	flowgo "github.com/onflow/flow-go/model/flow"
 	"github.com/rs/zerolog"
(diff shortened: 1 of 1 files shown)


bluesign:
@pattyshack thank you so much for `fvm.WithAuthorizationChecksEnabled` and `fvm.WithSequenceNumberCheckAndIncrementEnabled`

blockchain.go
```diff
@@ -1037,7 +1040,13 @@ func (b *Blockchain) GetAccountStorage(address sdk.Address) (*AccountStorage, er
...
 		WithMaxValueSizeAllowed(b.vmCtx.MaxStateValueSize)
 
-	txnPrograms := programs.NewEmptyPrograms()
+	derivedBlockData := derived.NewEmptyDerivedBlockData()
+	derivedTxnData, err := derivedBlockData.NewSnapshotReadDerivedTransactionData(
+		derived.EndOfBlockExecutionTime,
+		derived.EndOfBlockExecutionTime)
+	if err != nil {
+		return nil, err
+	}
```
bluesign, line 1049: nit, maybe some method like `derived.NewEmptyDerivedTransactionData()` in the future to fvm?

bluesign (approved):

blockchain.go
```diff
@@ -417,7 +417,10 @@ func configureFVM(conf config, blocks *blocks) (*fvm.VirtualMachine, fvm.Context
 	}
 
 	if !conf.TransactionValidationEnabled {
-		fvmOptions = append(fvmOptions, fvm.WithTransactionProcessors(fvm.NewTransactionInvoker()))
+		fvmOptions = append(
+			fvmOptions,
+			fvm.WithAuthorizationChecksEnabled(false),
```
sideninja, line 422: Should we also add `WithAccountKeyWeightThreshold` with negative value to skip signature verification? cc @bluesign

sideninja (approved):

blockchain.go
pattyshack: you're using an internal api here.  if you're using the fvm public interface, these are optional

bluesign: hmm I didn't know you can do this via fvm public interfaces. let me have a look and make a PR, thanks @pattyshack

pattyshack: i don't think there's an equivalent public interface.  i'm just pointing out that fvm is designed around the VM (Run) interface in general

bluesign: ah ok got it thanks, though probably I still workaround fvm here.

PR #315, opened 2023-02-07
Create config endpoint

### LINKED ISSUE #314: Create Configuration Endpoint
## Instructions
Create an endpoint for retrieving the configuration info of a running emulator.
This will be useful for integrations, such as the VSCode extension, to determine if an emulator can be connected to based on its service key or start up location.

DylanTinianov:
Closes #314 

## Description
Added /emulator/config endpoint to display configuration info for the ServiceKey

______

For contributor use:


server/config.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+)
+
+type ConfigInfo struct {
+	ServiceKey string `json:"service_key"`
+	StartupDir string `json:"startup_dir"`
```
sideninja, line 30: I don't think this is needed. If the service key is the same then it doesn't really matter where it was run.

server/config.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+var _ http.Handler = &ConfigInfo{}
+
+func NewConfigInfo(emulatorServer *EmulatorServer) *ConfigInfo {
+	startupDir, _ := os.Getwd()
+
+	return &ConfigInfo{
+		ServiceKey: fmt.Sprintf("0x%s", emulatorServer.blockchain.ServiceKey().Address.Hex()),
```
sideninja, line 39: This should return the public key which you can compare no the address. Also the value should better be `ServicePublicKey`

server/admin.go
```diff
@@ -34,6 +34,7 @@ import (
 const (
 	LivenessPath    = "/live"
 	MetricsPath     = "/metrics"
+	ConfigPath      = "/config"
```
sideninja, line 37: This is better if it's under `/emulator` as that contains more admin endpoints for the emulator. So the route would be `/emulator/config`

sideninja (commented):
Left comments

server/admin.go
DylanTinianov: Moved :)

server/config.go
DylanTinianov: Actually we need to check the `ServicePrivateKey` which is  stored in flow.json

server/config.go
DylanTinianov: Removed

server/emulator.go
```diff
@@ -62,6 +62,9 @@ func NewEmulatorAPIServer(server *EmulatorServer, backend *backend.Backend, stor
 
 	router.HandleFunc("/emulator/storages/{address}", r.Storage)
 
+	config := NewConfigInfo(server)
+	router.Handle("/emulator/config", config)
```
bluesign, line 66: maybe just basic `HandleFunc` ? and a new function there, instead new file `config.go` ?

server/config.go
```diff
@@ -0,0 +1,41 @@
...
+
+type ConfigInfo struct {
+	ServiceKey string `json:"service_key"`
+}
+
+var _ http.Handler = &ConfigInfo{}
+
+func NewConfigInfo(emulatorServer *EmulatorServer) *ConfigInfo {
+	return &ConfigInfo{
+		ServiceKey: emulatorServer.blockchain.ServiceKey().PrivateKey.String(),
```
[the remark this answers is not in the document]
sideninja: public yeah

server/emulator.go
sideninja: +1

sideninja: I would add to admin.go as we have those gathered there

server/config.go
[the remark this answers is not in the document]
DylanTinianov: The private key is what's exposed in the flow.json config, so we need to expose this to validate the emulator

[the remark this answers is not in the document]
sideninja: What he said. In vscode you will be able to derive public key from private key, I believe you can even use FCL for that.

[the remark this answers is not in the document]
sideninja: Although emulator is not meant to be used in production I can already see some weird case where someone will use it for some storage and then the private key will be leaked.

sideninja (approved):

server/config.go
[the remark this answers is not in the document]
DylanTinianov: Sounds good

PR #332, opened 2023-03-13
Expose Cadence Code Coverage through Flow Emulator

m-Peter:
**Closes** https://github.com/onflow/flow-emulator/issues/333
**Depends On** https://github.com/onflow/flow-go/pull/4126

## Description

 Allow the Flow Emulator to utilize the built-in code coverage functionality from Cadence, and expose it to developers through:
- API (programmatic)
- HTTP endpoint (visual)
______

For contributor use:

### DIFF
--- a/blockchain.go
+++ b/blockchain.go
@@ -1555,5 +1555,5 @@ func (b *Blockchain) CoverageReport() *runtime.CoverageReport {
 }
 
 func (b *Blockchain) ResetCoverageReport() {
-	b.coverageReportedRuntime.CoverageReport = runtime.NewCoverageReport()
+	b.coverageReportedRuntime.Reset()
 }
--- a/blockchain.go
+++ b/blockchain.go
@@ -1554,6 +1554,10 @@ func (b *Blockchain) CoverageReport() *runtime.CoverageReport {
 	return b.coverageReportedRuntime.CoverageReport
 }
 
+func (b *Blockchain) SetCoverageReport(coverageReport *runtime.CoverageReport) {
+	b.coverageReportedRuntime.CoverageReport = coverageReport
+}
+
 func (b *Blockchain) ResetCoverageReport() {
 	b.coverageReportedRuntime.Reset()
 }
--- a/blockchain.go
+++ b/blockchain.go
@@ -69,6 +69,8 @@ type Blockchain struct {
 	currentScriptID        string
 
 	conf config
+
+	coverageReportedRuntime *CoverageReportedRuntime
 }
 
 type ServiceKey struct {
@@ -151,6 +153,7 @@ type config struct {
 	Logger                       zerolog.Logger
 	TransactionValidationEnabled bool
 	ChainID                      flowgo.ChainID
+	CoverageReportingEnabled     bool
 }
 
 func (conf config) GetStore() storage.Store {
@@ -206,6 +209,7 @@ var defaultConfig = func() config {
 		Logger:                       zerolog.Nop(),
 		TransactionValidationEnabled: true,
 		ChainID:                      flowgo.Emulator,
+		CoverageReportingEnabled:     false,
 	}
 }()
 
@@ -373,6 +377,18 @@ func WithChainID(chainID flowgo.ChainID) Option {
 		c.ChainID = chainID
 	}
 }
+
+// WithCoverageReportingEnabled enables/disables Cadence code coverage reporting.
+//
+// If set to false, the emulator will not collect/expose coverage information for Cadence code.
+//
+// The default is false.
+func WithCoverageReportingEnabled(enabled bool) Option {
+	return func(c *config) {
+		c.CoverageReportingEnabled = enabled
+	}
+}
+
 func (b *Blockchain) ReloadBlockchain() error {
 	var err error
 
@@ -484,6 +500,24 @@ func (b *Blockchain) LoadSnapshot(name string) error {
 func configureFVM(blockchain *Blockchain, conf config, blocks *blocks) (*fvm.VirtualMachine, fvm.Context, error) {
 	vm := fvm.NewVirtualMachine()
 
+	config := runtime.Config{
+		Debugger:                 blockchain.debugger,
+		AccountLinkingEnabled:    true,
+		CoverageReportingEnabled: conf.CoverageReportingEnabled,
+	}
+	coverageReportedRuntime := &CoverageReportedRuntime{
+		Runtime:        runtime.NewInterpreterRuntime(config),
+		CoverageReport: runtime.NewCoverageReport(),
+		Environment:    runtime.NewBaseInterpreterEnvironment(config),
+	}
+	customRuntimePool := reusableRuntime.NewCustomReusableCadenceRuntimePool(
+		1,
+		config,
+		func(config runtime.Config) runtime.Runtime {
+			return coverageReportedRuntime
+		},
+	)
+
 	fvmOptions := []fvm.Option{
 		fvm.WithLogger(conf.Logger),
 		fvm.WithChain(conf.GetChainID().Chain()),
@@ -494,14 +528,7 @@ func configureFVM(blockchain *Blockchain, conf config, blocks *blocks) (*fvm.Vir
 		fvm.WithCadenceLogging(true),
 		fvm.WithAccountStorageLimit(conf.StorageLimitEnabled),
 		fvm.WithTransactionFeesEnabled(conf.TransactionFeesEnabled),
-		fvm.WithReusableCadenceRuntimePool(
-			reusableRuntime.NewReusableCadenceRuntimePool(
-				1,
-				runtime.Config{
-					Debugger:              blockchain.debugger,
-					AccountLinkingEnabled: true,
-				}),
-		),
+		fvm.WithReusableCadenceRuntimePool(customRuntimePool),
 	}
 
 	if !conf.TransactionValidationEnabled {
@@ -514,6 +541,8 @@ func configureFVM(blockchain *Blockchain, conf config, blocks *blocks) (*fvm.Vir
 	ctx := fvm.NewContext(
 		fvmOptions...,
 	)
+
+	blockchain.coverageReportedRuntime = coverageReportedRuntime
 
 	return vm, ctx, nil
 }
@@ -698,11 +727,6 @@ func (b *Blockchain) newFVMContextFromHeader(header *flowgo.Header) fvm.Context 
 	return fvm.NewContextFromParent(
 		b.vmCtx,
 		fvm.WithBlockHeader(header),
-		fvm.WithReusableCadenceRuntimePool(
-			reusableRuntime.NewReusableCadenceRuntimePool(
-				1,
-				runtime.Config{Debugger: b.debugger, AccountLinkingEnabled: true}),
-		),
 	)
 }
 
@@ -1524,4 +1548,12 @@ func (b *Blockchain) SetDebugger(debugger *interpreter.Debugger) {
 
 func (b *Blockchain) EndDebugging() {
 	b.SetDebugger(nil)
+}
+
+func (b *Blockchain) CoverageReport() *runtime.CoverageReport {
+	return b.coverageReportedRuntime.CoverageReport
+}
+
+func (b *Blockchain) ResetCoverageReport() {
+	b.coverageReportedRuntime.CoverageReport = runtime.NewCoverageReport()
 }
--- a/cmd/emulator/start/start.go
+++ b/cmd/emulator/start/start.go
@@ -73,6 +73,7 @@ type Config struct {
 	ChainID                   string        `default:"emulator" flag:"chain-id" info:"chain to emulate for address generation. Valid values are: 'emulator', 'testnet', 'mainnet'"`
 	RedisURL                  string        `default:"" flag:"redis-url" info:"redis-server URL for persisting redis storage backend ( redis://[[username:]password@]host[:port][/database] ) "`
 	SqliteURL                 string        `default:"" flag:"sqlite-url" info:"sqlite db URL for persisting sqlite storage backend "`
+	CoverageReportingEnabled  bool          `default:"false" flag:"coverage-reporting" info:"enable Cadence code coverage reporting"`
 }
 
 const EnvPrefix = "FLOW"
@@ -191,6 +192,7 @@ func Cmd(getServiceKey serviceKeyFunc) *cobra.Command {
 				RedisURL:                  conf.RedisURL,
 				ContractRemovalEnabled:    conf.ContractRemovalEnabled,
 				SqliteURL:                 conf.SqliteURL,
+				CoverageReportingEnabled:  conf.CoverageReportingEnabled,
 			}
 
 			emu := server.NewEmulatorServer(logger, serverConf)
--- a/coverage_reported_runtime.go
+++ b/coverage_reported_runtime.go
@@ -1,0 +1,172 @@
+/*
+ * Flow Emulator
+ *
+ * Copyright 2019 Dapper Labs, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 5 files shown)


blockchain.go
```diff
@@ -612,11 +619,6 @@ func (b *Blockchain) newFVMContextFromHeader(header *flowgo.Header) fvm.Context
 	return fvm.NewContextFromParent(
 		b.vmCtx,
 		fvm.WithBlockHeader(header),
-		fvm.WithReusableCadenceRuntimePool(
-			reusableRuntime.NewReusableCadenceRuntimePool(
-				1,
-				runtime.Config{Debugger: b.debugger, AccountLinkingEnabled: true}),
-		),
```
bluesign, line 619: is there a reason to remove this? this should be needed for debugger to function ( not sure though, flow-go is changing too fast )

server/emulator.go
```diff
@@ -65,6 +65,8 @@ func NewEmulatorAPIServer(server *EmulatorServer, backend *backend.Backend, stor
 
 	router.HandleFunc("/emulator/config", r.Config)
 
+	router.HandleFunc("/emulator/codeCoverage", r.CodeCoverage).Methods("GET")
+
 	return r
```
bluesign, line 70: nit: would it be possible to also add a method to reset coverage report ? It can be very useful.

blockchain.go
```diff
@@ -207,6 +210,7 @@ var defaultConfig = func() config {
 		Logger:                       zerolog.Nop(),
 		TransactionValidationEnabled: true,
 		ChainID:                      flowgo.Emulator,
+		CoverageReportingEnabled:     true,
 	}
```
bluesign, line 214: I think we can get this from command line flag ( default can be true ) I am also totally fine with making it default on without any flag. but if we make config I think making it configurable would be better

bluesign:
great work @m-Peter, left few comments

blockchain.go
```diff
@@ -445,6 +450,8 @@ func configureFVM(blockchain *Blockchain, conf config, blocks *blocks) (*fvm.Vir
 		fvmOptions...,
 	)
 
+	blockchain.coverageReport = ctx.ReusableCadenceRuntimePool.CoverageReport
+
```
bluesign, line 454: `./blockchain.go:453:61: ctx.ReusableCadenceRuntimePool.CoverageReport undefined (type "github.com/onflow/flow-go/fvm/runtime".ReusableCadenceRuntimePool has no field or method CoverageReport)`

bluesign (changes requested):

server/emulator.go
m-Peter: That is quite a nice idea :100: :100: :100: 
I will add it for sure!!!

blockchain.go
m-Peter: Yes, I definitely have to add a command line flag also. But I guess we need this one also, in case the `Blockchain` is utilized programmatically, for example, from another library. In this case the command line flag will not be sufficient I think. Correct me if I'm wrong though.

blockchain.go
m-Peter: I will have to bump the `flow-go` version, with a release that includes the changes from here: https://github.com/onflow/flow-go/pull/4038

blockchain.go
m-Peter: I had to remove this line here, because it would create a new `CoverateReport` object, which would be ephemeral during script & transaction execution.

The `ReusableCadenceRuntimePool` is properly initialized here: https://github.com/onflow/flow-emulator/blob/master/blockchain.go#L427-L446, so with:

```go
return fvm.NewContextFromParent(
    b.vmCtx,
    fvm.WithBlockHeader(header),
)
```

the `b.vmCtx` contains all the necessary config for the `Debugger`, `AccountLinkingEnabled` and `CoverageReportingEnabled` from the initial setup.

blockchain.go
bluesign: true, you can expose it with a function like `WithCoverageReportingEnabled`  ( example: https://github.com/onflow/flow-emulator/blob/1fd1c61822d4771ac2dcad7a4f908bd0586ac293/blockchain.go#L368 ) 

then you need to add to `configureBlockchain`: https://github.com/onflow/flow-emulator/blob/1fd1c61822d4771ac2dcad7a4f908bd0586ac293/server/server.go#L335

then to this `config`: https://github.com/onflow/flow-emulator/blob/1fd1c61822d4771ac2dcad7a4f908bd0586ac293/server/server.go#L85

and finally: https://github.com/onflow/flow-emulator/blob/1fd1c61822d4771ac2dcad7a4f908bd0586ac293/cmd/emulator/start/start.go#L42

and 

https://github.com/onflow/flow-emulator/blob/1fd1c61822d4771ac2dcad7a4f908bd0586ac293/cmd/emulator/start/start.go#L161

Turtles all the way down :)

blockchain.go
bluesign: ah you have right, I missed that

server/emulator.go
m-Peter: @bluesign With the current API, we can achieve this with something like:

```go
func (b *Blockchain) ResetCoverageReport() {
	b.vmCtx.ReusableCadenceRuntimePool.CoverageReport = runtime.NewCoverageReport()
	b.coverageReport = b.vmCtx.ReusableCadenceRuntimePool.CoverageReport
}
```

I am not a big fan of reaching this with `b.vmCtx.ReusableCadenceRuntimePool.CoverageReport`, however, the `CoverageReport` lacks a `Reset` method, for the time being.

What are your thoughts on this? Do we first add the required API in `CoverageReport`, or we add a workaround like above, and come back to fix it later on?

blockchain.go
m-Peter: Flags, flags everywhere :joy_cat: 
Updated in https://github.com/onflow/flow-emulator/pull/332/commits/8a6d81abc55a31654b1819e8391cec658e2e2af8

server/emulator.go
m-Peter: I have added here (https://github.com/onflow/flow-emulator/pull/332/commits/4e7b422e36c101f156ae74efa40ca262bd0958f6) the workaround, for reference.

m-Peter:
@turbolent I have added a commit (https://github.com/onflow/flow-emulator/pull/332/commits/6a46c114c8aeea46458c580ba7c4082419100917) which attempts to apply the suggestion from https://github.com/onflow/flow-go/pull/4038. It does not work as expected though, I guess I am missing something in the implementation.

Screenshot from 2023-03-16 13-32-58

The contract appears on the coverage report, only after the relevant script is executed. And for some reason, the declarations inside the `init()` method, have not been correctly recorded.

Screenshot from 2023-03-16 13-37-43

With the previous implementation, just by deploying the contract, it would immediately appear in the coverage report. And the declarations inside the `init()` method, have been correctly recorded.

Do you have any clue as to why this happens? :pray:

blockchain.go
```diff
@@ -207,6 +210,7 @@ var defaultConfig = func() config {
 		Logger:                       zerolog.Nop(),
 		TransactionValidationEnabled: true,
 		ChainID:                      flowgo.Emulator,
+		CoverageReportingEnabled:     true,
```
turbolent, line 213: Should this be enabled by default?

turbolent (approved):

blockchain.go
m-Peter: Not really, it's up for discussion. On another topic, have you had a look on this https://github.com/onflow/flow-emulator/pull/332#issuecomment-1471840314? I am struggling to find the cause of this, so we can contain the changes in this repository, and discard the `flow-go` companion PR.

cmd/emulator/start/start.go
```diff
@@ -73,6 +73,7 @@ type Config struct {
 	ChainID                   string        `default:"emulator" flag:"chain-id" info:"chain to emulate for address generation. Valid values are: 'emulator', 'testnet', 'mainnet'"`
 	RedisURL                  string        `default:"" flag:"redis-url" info:"redis-server URL for persisting redis storage backend ( redis://[[username:]password@]host[:port][/database] ) "`
 	SqliteURL                 string        `default:"" flag:"sqlite-url" info:"sqlite db URL for persisting sqlite storage backend "`
+	CoverageReportingEnabled  bool          `default:"true" flag:"coverage-reporting" info:"enable Cadence code coverage reporting"`
```
sideninja, line 76: probably better to not be enabled by default?

bluesign:
@m-Peter  can be related to reusable runtime fix we talked about.

coverage_reported_runtime.go
```diff
@@ -0,0 +1,172 @@
...
+
+func (crt *CoverageReportedRuntime) SetFvmEnvironment(
+	fvmEnv reusableRuntime.Environment,
+) {
+	crt.fvmEnv = fvmEnv
+}
+
+func (crt CoverageReportedRuntime) NewScriptExecutor(
+	script runtime.Script,
+	context runtime.Context,
```
sideninja, line 46: context should be the first argument

cmd/emulator/start/start.go
sideninja: why resolved?

sideninja (approved):

sideninja:
@m-Peter should we get this merged?

cmd/emulator/start/start.go
bluesign: sorry I misclicked on this probably

m-Peter: Good point, I will change the default behavior :+1:

blockchain.go
m-Peter: Updated default value to `false`

cmd/emulator/start/start.go
m-Peter: Updated the default value to `false`

m-Peter:
> @m-Peter can be related to reusable runtime fix we talked about.

@bluesign I tried this as well, but the same result, it only collects code coverage reports, upon script/transaction execution. It also misses some parts which are collected upon contract deployment. I am out of ideas on why this is the case, with a custom reusable Cadence runtime pool. I have applied the suggested comments, on the companion PR (https://github.com/onflow/flow-go/pull/4038).

m-Peter:
@bluesign Appreciate the help :pray: In this https://github.com/onflow/flow-emulator/pull/332#issuecomment-1471840314 you can see some screenshots with the exact problem.

m-Peter:
Yes, as the state of the two PRs, it works as expected. I edited my previous comment, sorry for that :sweat_smile: 

> Note: The problem appears only when we try to contain the changes in this repository, as pattyshack suggested here (https://github.com/onflow/flow-go/pull/4038#discussion_r1137460821). With the companion PR ihttps://github.com/https://github.com/onflow/flow-go/pull/4038, things work as expected.

bluesign:
@m-Peter yep it turned out that empty config related. ( the fix we discussed before )

bluesign:
```go
config := runtime.Config{
		Debugger:                 blockchain.debugger,
		AccountLinkingEnabled:    true,
		CoverageReportingEnabled: conf.CoverageReportingEnabled,
	}
	coverageReportedRuntime := CoverageReportedRuntime{
		Runtime:        runtime.NewInterpreterRuntime(config),
		CoverageReport: runtime.NewCoverageReport(),
		Environment:    runtime.NewBaseInterpreterEnvironment(config),
	}

	customRuntimePool := reusableRuntime.NewCustomReusableCadenceRuntimePool(
		10,
		func(config runtime.Config) runtime.Runtime {
			return coverageReportedRuntime
		},
	)

	for i := 0; i < 10; i++ {
		customRuntimePool.Return(reusableRuntime.NewReusableCadenceRuntime(coverageReportedRuntime, config))
	}
```

PS: not proud of this code :D

m-Peter:
@bluesign Hmm, I was sure that I tried it with the fix from Bastian, since I used the `master` branch. Let me check again, most likely I missed something.

bluesign:
@m-Peter seems it is not fixed the custom problem.

m-Peter:
My sincere apologies for the numerous `force-push` operations :pray:
After some back and forth with @bluesign, and his excellent debugging skills, I have brought this to what I consider a final version, ready for review. The prerequisite is the PR https://github.com/onflow/flow-go/pull/4126.

sideninja (approved):

sideninja (changes requested):
https://github.com/onflow/flow-go/pull/4126

bluesign:
@m-Peter I am not sure how I feel about [coverage_reported_runtime.go](https://github.com/onflow/flow-emulator/pull/332#diff-a3eb5c3acfb7ec0fb1de7cc95f8750ee021315c8324a875c6a2a6f652b827021) I think @janezpodhostnik and @turbolent knows better but it feels like a bit off to me. 

I feel there should be another way to push coverageReport ? with flow-go fix do we still need this?

m-Peter:
@bluesign I personally agree with you, having to create a runtime wrapper, for the sake of coverage report, seems like an extra place where we need to conform with the runtime interface. However, this was suggested from the flow-go team, as an alternative that doesn't need any changes besides this repository. Since you uncovered the bug of config swallow, it turns out that it needs at least that. That's why I have closed the previous companion PR of flow-go (https://github.com/onflow/flow-go/pull/4038). Personally I like the approach that Janez suggested, so that we don't create a runtime wrapper, but we add the method to inject a coverage report in the reusable cadence runtime pool. How do you feel about this alternative?

bluesign:
I think it would be best to pass it as runtime.Config:

```go
runtime.Config{
		Debugger:                 blockchain.debugger,
		AccountLinkingEnabled:    true,
		CoverageReportingEnabled: conf.CoverageReportingEnabled,
                CoverageReport: blockchain.coverageReport
}),
```

with config fix, it should be enough without touching the flow-go.  



Btw totally off topic but it would be nice to expose CoverageReport via interface, something like: 

```go
type LineHitReporter interface {
  AddLineHit(location Location, line int) 
}
```

so people can add different coverage modes. But actually what would be better is to put `OnStatement` handler to runtime config. https://github.com/onflow/cadence/blob/5bfce97a565740cd0257311a4bb339e4c258b75a/runtime/environment.go#LL147C3-L147C14

This way coverage report doesn't need to use this : https://github.com/onflow/cadence/blob/5bfce97a565740cd0257311a4bb339e4c258b75a/runtime/environment.go#L608

can take this function to coverageReport and hook it to OnStatement handler. 
https://github.com/onflow/cadence/blob/5bfce97a565740cd0257311a4bb339e4c258b75a/runtime/environment.go#L613-L622

sideninja:
@m-Peter what is the status of this PR?

m-Peter:
@sideninja The prerequisite PR was merged yesterday (https://github.com/onflow/flow-go/pull/4126), and with it in a released version of `flow-go` we are ready to merge :pray: . Nothing else to add on my side.

m-Peter:
I did a rebase on master, to get the changes from https://github.com/onflow/flow-emulator/pull/347. It includes the prerequisite PR.

sideninja (approved):

sideninja:
@bluesign can you re-review?

bluesign:
@m-Peter I was gonna review but shouldn't we update this to use https://github.com/onflow/flow-emulator/pull/332 ?

m-Peter:
@bluesign I think we can first make this available to the developer community, and we can follow up for this change. As it is, it works, and only the internal implementation will change. It will take some time to make all the releases available and update the dependencies.. What do you think?

bluesign:
@m-Peter yeah but I think we at least need flow-go PR to merged here ( custom pool fix ) otherwise it breaks the debugger probably.

m-Peter:
@bluesign Yes, it was released, and updated on this repository. I have rebased to include it. But if it's not much trouble, you can double check that nothing was broken.

bluesign:
> @bluesign Yes, it was released, and updated on this repository. I have rebased to include it. But if it's not much trouble, you can double check that nothing was broken.

sure I will have a look

bluesign:
@m-Peter needs rebase to master ( probably on the way you will also need to update the flow-go version, as we don't have the patched one currently it seems )

m-Peter:
@bluesign I have rebased after pulling the latest changes from master. It seems to me that we do have the fix though, otherwise, this piece of [code](https://github.com/onflow/flow-emulator/pull/332/files#diff-1fb2891d150a66e1319495e4f10a7ce04855476836a3a6b17f253e312f37d23bR513-R519), would have compilation errors. Any clue?

bluesign (approved):

bluesign:
seems all good after rebase, I think we can merge this, then after cadence changes, we can come back to this.

turbolent (approved):

PR #351, opened 2023-04-08
Convert Cadence logs to realtime

### LINKED ISSUE #302: Real-time log from Cadence
### Issue To Be Solved

Currently logs are shown after transaction or script execution. We can make them real-time, by using fvm cadence logging. 

I think after #293 , shouldn't be so hard. ( We also need for debugger )

bluesign:
Closes #302 

## Description

Converts cadence logs to realtime output instead of after script/transaction execution. 

______

For contributor use:

### DIFF
--- a/blockchain.go
+++ b/blockchain.go
@@ -14,6 +14,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/logrusorgru/aurora"
+	"strings"
 	"sync"
 	"time"
 
@@ -149,6 +151,7 @@ type config struct {
 	MinimumStorageReservation    cadence.UFix64
 	StorageMBPerFLOW             cadence.UFix64
 	Logger                       zerolog.Logger
+	ServerLogger                 zerolog.Logger
 	TransactionValidationEnabled bool
 	ChainID                      flowgo.ChainID
 }
@@ -204,6 +207,7 @@ var defaultConfig = func() config {
 		TransactionExpiry:            0, // TODO: replace with sensible default
 		StorageLimitEnabled:          true,
 		Logger:                       zerolog.Nop(),
+		ServerLogger:                 zerolog.Nop(),
 		TransactionValidationEnabled: true,
 		ChainID:                      flowgo.Emulator,
 	}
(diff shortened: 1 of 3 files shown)


blockchain.go
```diff
@@ -698,11 +727,6 @@ func (b *Blockchain) newFVMContextFromHeader(header *flowgo.Header) fvm.Context
 	return fvm.NewContextFromParent(
 		b.vmCtx,
 		fvm.WithBlockHeader(header),
-		fvm.WithReusableCadenceRuntimePool(
-			reusableRuntime.NewReusableCadenceRuntimePool(
-				1,
-				runtime.Config{Debugger: b.debugger, AccountLinkingEnabled: true}),
-		),
```
turbolent, line 705: Was this removal intentional? Is it not needed anymore?

turbolent (commented):
Nice! Great idea 👌

blockchain.go
bluesign: yeah, @m-peter pointed out[0] in coverage PR that it is not needed anymore as config will be passed via parent context.

[0] https://github.com/onflow/flow-emulator/pull/332#discussion_r1135265180

m-Peter: Exactly, I observed that it was redundant for the `Debugger` & `AccountLinkingEnabled` config values. Moreover, the removal was necessary for the coverage reporting feature to work properly. However, at the time of writing that comment, the `ReusableCadenceRuntimePool` was responsible for also initializing the `CoverageReport` object. This was later changed. If there's no specific reason for constructing a new `ReusableCadenceRuntimePool`, then I would say that it's safe to remove.

turbolent: Sounds good, thanks for the explanation! I was just wondering how the removal relates to this PR

turbolent (approved):

blockchain.go
```diff
@@ -481,11 +494,27 @@ func (b *Blockchain) LoadSnapshot(name string) error {
...
 
+type CadenceHook struct {
+	MainLogger *zerolog.Logger
+}
+
+func (h CadenceHook) Run(e *zerolog.Event, level zerolog.Level, msg string) {
+	if level != zerolog.NoLevel && strings.HasPrefix(msg, "Cadence log:") {
+		h.MainLogger.Info().Msg(
+			strings.Replace(msg,
+				"Cadence log:",
```
turbolent, line 505:
```suggestion
	const logPrefix = "Cadence log:"
	if level != zerolog.NoLevel && strings.HasPrefix(msg, logPrefix) {
		h.MainLogger.Info().Msg(
			strings.Replace(msg,
				logPrefix,
```

blockchain.go
bluesign: ah yeah sorry, actually it was something left from when I tested if removing was safe, then it somehow came into this PR unintentionally, but then I thought adding would not make sense, for the sake of it. Sorry created a bit confusion

PR #367, opened 2023-05-05
Expose deployment of common contracts to `Blockchain` type

m-Peter:
Work towards: https://github.com/onflow/developer-grants/issues/148

## Description

We introduce a new option, called `Contracts`, in order to programmatically specify which contracts should a `Blockchain` deploy (this is mostly for the common contracts, it does not affect the system contracts, which are essential to the function of the emulator/blockchain).

The end goal is to be able to utilize it in other libraries, for example the `cadence-tools/test` repo, with something like:

```go
blockchain = newBlockchain(
    emulator.WithCoverageReportingEnabled(true),
    emulator.Contracts(emulator.CommonContracts),
)
```

This will enable the Cadence Testing Framework to have the common contracts available by default. e.g: `NonFungibleToken` / `MetadataViews` / `FUSD` / `ExampleNFT` / `NFTStorefront` /  `NFTStorefrontV2` etc.

______

For contributor use:


blockchain.go
```diff
@@ -449,8 +459,18 @@ func NewBlockchain(opts ...Option) (*Blockchain, error) {
...
 	}
-	return b, nil
+	if conf.WithContracts {
+		deployments, err := DeployContracts(b)
+		if err != nil {
+			conf.ServerLogger.Error().Err(err).Msg("❗  Failed to deploy contracts")
+		}
 
+		for _, contract := range deployments {
+			conf.ServerLogger.Info().Fields(map[string]any{
```
m-Peter, line 469: Two side effects of the logging taking place inside the `Blockchain`, are the following:

```bash
  flow-code-coverage git:(main) ✗ ~/Dev/forks/flow-cli/cmd/flow/flow-x86_64-linux- emulator --coverage-reporting --contracts       
5:30PM INF ⚙️   Using service account 0xf8d6e0586b0a20c7 serviceAddress=f8d6e0586b0a20c7 serviceHashAlgo=SHA3_256 servicePrivKey=2619878f0e2ff438d17835c2a4561cb87b4d24d72d12ec34569acd0dd4af7c21 servicePubKey=ebd3b08b2926ae27605f6ab060af8e0efd53f638a2664c55bdc76080ae0d2ad42b3a2835dce6059a54702df26296b24b70c575fa69f965228542dc4349fa6c04 serviceSigAlgo=ECDSA_P256
5:30PM INF 💵  FUSD contract                             FUSD=0xf8d6e0586b0a20c7
5:30PM INF ✨  NFT contract                              NonFungibleToken=0xf8d6e0586b0a20c7
5:30PM INF ✨  Metadata views contract                   MetadataViews=0xf8d6e0586b0a20c7
5:30PM INF ✨  Example NFT contract                      ExampleNFT=0xf8d6e0586b0a20c7
5:30PM INF ✨  NFT Storefront contract v2                NFTStorefrontV2=0xf8d6e0586b0a20c7
5:30PM INF ✨  NFT Storefront contract                   NFTStorefront=0xf8d6e0586b0a20c7
5:30PM INF 📜  Flow contract                             FlowStorageFees=0xf8d6e0586b0a20c7
5:30PM INF 📜  Flow contract                             FlowServiceAccount=0xf8d6e0586b0a20c7
5:30PM INF 📜  Flow contract                             FlowToken=0x0ae53cb6e3f42a79
5:30PM INF 📜  Flow contract                             FungibleToken=0xee82856bf20e2aa6
5:30PM INF 📜  Flow contract                             FlowFees=0xe5a8b7f23e8b548f
```
The common contracts are printed before the system contracts.

And second, the deployment info logs, are printed in the `test` command:

```bash
➜  flow-code-coverage git:(main) ✗ ~/Dev/forks/flow-cli/cmd/flow/flow-x86_64-linux- test --cover test_string_utils.cdc
Running tests...⠇
5:31PM INF 💵  FUSD contract FUSD=0xf8d6e0586b0a20c7
5:31PM INF ✨  NFT contract NonFungibleToken=0xf8d6e0586b0a20c7
5:31PM INF ✨  Metadata views contract MetadataViews=0xf8d6e0586b0a20c7
5:31PM INF ✨  Example NFT contract ExampleNFT=0xf8d6e0586b0a20c7
5:31PM INF ✨  NFT Storefront contract v2 NFTStorefrontV2=0xf8d6e0586b0a20c7

Test results: "test_string_utils.cdc"
- PASS: testFormat
- PASS: testExplode
- PASS: testTrimLeft
- PASS: testTrim
- PASS: testReplaceAll
- PASS: testHasPrefix
- PASS: testHasSuffix
- PASS: testIndex
- PASS: testCount
- PASS: testContains
- PASS: testSubstringUntil
- PASS: testSplit
- PASS: testJoin
Coverage: 84.4% of statements
```

I do not quite like that :sweat_smile:

bluesign:
I think we need a different injection system for this. 

I think blockchain should not know about contracts but should get them as parameter somehow. Ideally we would have fvm bootstrap those. We should have an issue on that.

m-Peter:
Thanks for the comment @bluesign :pray: 
My first approach, usually revolves around at keeping the changes contained, that's why I just refactored the existing functionality. If the team behind `flow-go` is open to such an addition, I could give it a shot.
Another approach, is to just expose publicly the common contract deployment logic (`contracts.go` file), so that `cadence-tools/test` package could utilize it.
What do you think about this?

bluesign:
I think first step we can send them via struct at first ( like DeployDescription here ) to blockchain. emulator.WithContracts can took `[]DeployDescription` etc. 

Then maybe we can bump the issue: https://github.com/onflow/flow-go/issues/2812 If fvm can expose us contracts in bootstrap, it maybe better than to deploy them afterwards. ( So we will have them all deployed even when block height is zero )

But I think we can merge this if we just change it to `[]DeployDescription`, so people can bootstrap via overflow, flowkit etc ( we can even expose default contracts.go

m-Peter:
Perfect, much thanks for the detailed explanation :bow: 
I will make the changes, and ping you :pray:

blockchain.go
m-Peter: Update: This is all resolved after the refactoring suggested by @bluesign

blockchain.go
```diff
@@ -158,6 +157,7 @@ type config struct {
 	TransactionValidationEnabled bool
 	ChainID                      flowgo.ChainID
 	CoverageReportingEnabled     bool
+	WithContracts                []ContractDescription
```
sideninja, line 160: Maybe drop the With prefix, or rename to something like InitialContracts, or StartupContracts, or just Contracts

blockchain.go
```diff
@@ -449,6 +461,12 @@ func NewBlockchain(opts ...Option) (*Blockchain, error) {
 	if err != nil {
 		return nil, err
 	}
+	if conf.WithContracts != nil && len(conf.WithContracts) > 0 {
```
sideninja, line 464: You can just check for len(), because if nil it would just be 0.

blockchain.go
```diff
@@ -943,7 +961,7 @@ func (b *Blockchain) GetTransactionResult(ID sdk.Identifier) (*sdk.TransactionRe
 // GetAccountByIndex returns the account for the given address.
 func (b *Blockchain) GetAccountByIndex(index uint) (*sdk.Account, error) {
 
-	generator := flow.NewAddressGenerator(sdk.ChainID(b.vmCtx.Chain.ChainID()))
+	generator := sdk.NewAddressGenerator(sdk.ChainID(b.vmCtx.Chain.ChainID()))
```
sideninja, line 964: why this changes?

contracts.go
```diff
@@ -0,0 +1,127 @@
...
+	flowsdk "github.com/onflow/flow-go-sdk"
+	"github.com/onflow/flow-go-sdk/templates"
+	"github.com/onflow/flow-go/fvm"
+	flowgo "github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow-nft/lib/go/contracts"
+	fusd "github.com/onflow/fusd/lib/go/contracts"
+	nftstorefront "github.com/onflow/nft-storefront/lib/go/contracts"
+)
+
+var CommonContracts = func() []ContractDescription {
```
sideninja, line 15: NIT: I think this could be a variable not a function. But I'm not sure if better since then you could modify it.

sideninja (approved):
Left some minor comments, otherwise looks great, thank you!

blockchain.go
m-Peter: Both flow and sdk were actually the same package imported twice, and the editor displayed a linting error. So I kept only one of the imports. Did I misunderstood this one?

blockchain.go
m-Peter: Ah, good suggestion 🙌

contracts.go
m-Peter: I am new to Golang. Could I have a global variable definition, with a body? Because I need the chain and service address intermediate variables.

blockchain.go
sideninja: oh, nice.

blockchain.go
m-Peter: Updated in https://github.com/onflow/flow-emulator/pull/367/commits/b4ab4b80457b775c2c360df6bdc8cfdf98013044

blockchain.go
m-Peter: Fixed in https://github.com/onflow/flow-emulator/pull/367/commits/b4ab4b80457b775c2c360df6bdc8cfdf98013044

contracts.go
m-Peter: I personally like the function version, because all the intermediate variables, are not exposed to the rest of the file :innocent:

m-Peter:
cc @turbolent @SupunS 

This is the necessary ground-work, for the [Deployed Contracts](https://github.com/onflow/cadence/issues/1889) improvement.
It will allow integration tests to directly import the core contracts from the service account address, e.g:

```cadence
import ArrayUtils from "ArrayUtils.cdc"
import FUSD from 0xf8d6e0586b0a20c7

pub contract StringUtils {
    ...
}
```

Once this is merged, I will open up a follow-up PR on the `cadence-tools/test` repository, with the necessary changes. To provide better flexibility, we might also bootstrap the configuration mapping:

```cadence
blockchain.useConfiguration(Test.Configuration({
    "NonFungibleToken": blockchain.serviceAccount().address,
    ...
}))
```

Exposing the `blockchain.serviceAccount()`, will also be quite essential, in order to transfer FLOW tokens to other accounts, from the service account.

sideninja (approved):

sideninja:
@m-Peter good to merge?

m-Peter:
@sideninja I think yes, we can proceed with merging this :pray:

PR #377, opened 2023-05-16
Remote storage

### LINKED ISSUE #370: Implement remote storage
Implement a storage that uses the archive node to get the registers. 
https://github.com/onflow/flow-archive

Implementing the store interface defined here https://github.com/onflow/flow-emulator/blob/8ec65fa00f3c00d5ce2aec156de1d74369f183ad/storage/store.go#L123 
We can reuse most of the default store implementation but swap out the LedgerByHeight to return the storage snapshot provided by archive node. 
The first step is to implement this for one-off transactions only, the second step would be to keep changes local and then fork the network state by first trying to see if we have local changes for a read and if not read from remote.

sideninja:
Closes #370 

## Description
This PR implements remote storage, which enables you to run transactions against testnet/mainnet and have a local fork of the chain.

______

For contributor use:


server/grpc.go
```diff
@@ -20,11 +20,11 @@ package server
 
 import (
 	"fmt"
+	"github.com/onflow/flow-go/access"
```
janezpodhostnik, line 23: linting issue?

server/grpc.go
```diff
@@ -63,13 +63,11 @@ func NewGRPCServer(logger *zerolog.Logger, b *backend.Backend, chain flow.Chain,
...
 
 	legacyaccessproto.RegisterAccessAPIServer(grpcServer, legacyaccess.NewHandler(adaptedBackend, chain))
-	accessproto.RegisterAccessAPIServer(
-		grpcServer,
-		access.NewHandler(
-			adaptedBackend,
-			chain,
-			mockHeaderCache{},
-			me))
+	accessproto.RegisterAccessAPIServer(grpcServer, access.NewHandler(
```
janezpodhostnik, line 66: I cant spot the change here besides the formatting....

storage/remote/store.go
```diff
@@ -0,0 +1,141 @@
...
+	if !ok {
+		return nil, fmt.Errorf("chain %s not supported with remote store", chainID.String())
+	}
+
+	conn, err := grpc.Dial(
+		host,
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not connect to archive node")
```
janezpodhostnik, line 42: just use `fmt.Errorf("...%w",err)` I think

storage/remote/store.go
```diff
@@ -0,0 +1,141 @@
...
+		return nil, err
+	}
+
+	store := &Store{
+		client: archive.NewAPIClient(conn),
+		Store:  memorySql,
+	}
+
+	store.DataGetter = store
+	store.DataSetter = store
```
janezpodhostnik, line 56: duplicated line

storage/sqlite/store.go
```diff
@@ -52,6 +54,54 @@ type Store struct {
...
+		return nil, err
+	}
+
+	store = &Store{
+		db:  db,
+		url: url,
+	}
+
+	store.DataSetter = store
+	store.DataGetter = store
```
janezpodhostnik, line 87: duplicated line

storage/remote/store.go
```diff
@@ -0,0 +1,141 @@
...
+	block, err := s.BlockByHeight(ctx, heightRes.Height)
+	if err != nil {
+		return flowgo.Block{}, err
+	}
+
+	return *block, nil
+}
+
+func (s *Store) BlockByHeight(ctx context.Context, height uint64) (*flowgo.Block, error) {
+	// todo use local storage first as a cache
```
janezpodhostnik, line 88: Will this be a separate PR?

Without this you cannot make changes with local transactions, right?

janezpodhostnik (commented):
Looks good to me, just some minor questions and comments.

storage/remote/store.go
sideninja: one is Set-er one Get-er?

server/grpc.go
sideninja: Yes, will fix

storage/remote/store.go
sideninja: You can because this function is not used for that, the get ledger one is which first checks local. I would make it separate PR just to keep it small.

sideninja:
> Looks good to me, just some minor questions and comments.

Addressed

storage/remote/store.go
```diff
@@ -0,0 +1,144 @@
+package remote
```
turbolent, line 1: License headers missing

storage/remote/store.go
```diff
@@ -0,0 +1,144 @@
...
+
+type Store struct {
+	*sqlite.Store
+	client archive.APIClient
+}
+
+func New(chainID flowgo.ChainID) (*Store, error) {
+	archiveHosts := map[flowgo.ChainID]string{
+		flowgo.Mainnet: "archive.mainnet.nodes.onflow.org:9000",
+		flowgo.Testnet: "archive.testnet.nodes.onflow.org:9000",
```
turbolent, line 29: It might make sense to make the archive host a configuration option and use these values as defaults

storage/remote/store.go
```diff
@@ -0,0 +1,144 @@
...
+		Payload: &payload,
+		Header:  &header,
+	}, nil
+}
+
+func (s *Store) LedgerByHeight(
+	ctx context.Context,
+	blockHeight uint64,
+) state.StorageSnapshot {
+	_ = s.SetBlockHeight(blockHeight)
```
turbolent, line 112: Maybe handle the error here instead of ignoring it

storage/remote/store.go
```diff
@@ -0,0 +1,144 @@
...
+
+	return snapshot.NewReadFuncStorageSnapshot(func(id flowgo.RegisterID) (flowgo.RegisterValue, error) {
+		// first try to see if we have local stored ledger
+		value, err := s.DefaultStore.GetBytesAtVersion(ctx, "ledger", []byte(id.String()), blockHeight)
```
turbolent, line 116: Maybe refactor the magic `"ledger"` value into a constant.

Where is it used for writing?

transaction_test.go
```diff
@@ -1902,3 +1903,120 @@ func TestRollbackTransaction(t *testing.T) {
 	IncrementHelper(t, b, counterAddress, addTwoScript, 2)
 
 }
+
+func Test_SimulatedMainnetTransaction(t *testing.T) {
```
turbolent, line 1907: Are these tests actually connecting to the real archive nodes? It might make sense to mock/stub this away, to avoid depending on such a remote system and avoid flakiness when there are connection or availability issues

turbolent (commented):
Awesome! This feature is going to be incredibly useful, great job!

storage/remote/store.go
sideninja: Yep

storage/remote/store.go
sideninja: The writing part is in another package and I don't like exposing it, but yeah I can make it a constant.

transaction_test.go
sideninja: Already experienced some flakiness so yeah, good idea for sure.

sideninja: But will have to look into how hard it is to mock all the requests, as there are so many I believe.

storage/remote/store.go
bluesign: maybe we can move this to constructor also

storage/remote/store.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/onflow/flow-go/ledger/complete"
+	flowgo "github.com/onflow/flow-go/model/flow"
+	"github.com/pkg/errors"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+type Store struct {
+	*sqlite.Store
+	client archive.APIClient
```
bluesign, line 23: maybe we can add fork height here, then we can split lookups with height below

bluesign: maybe can need to commit a block to fork and makes things a bit easier.

sideninja:
self note: add closer

storage/remote/store.go
sideninja: I tried implementing this, but it seems there's a problem with read data, there's some data that is requested in the after fork heights that wasn't yet saved locally, so I believe I can't make this assumption.

transaction_test.go
sideninja: > Are these tests actually connecting to the real archive nodes? It might make sense to mock/stub this away, to avoid depending on such a remote system and avoid flakiness when there are connection or availability issues

@turbolent I mocked the Archive client and used it to test but it's proving very hard to write a good test since there are a lot of register values being requested and transferred. I'm not sure is there's a better way, so let me know if you have an idea. 
Otherwise I excluded this test from transaction tests and moved them to separate test file with `integration` tag.

sideninja:
Can I get another review on this @janezpodhostnik @turbolent @bluesign

README.md
```diff
@@ -213,6 +213,15 @@ To generate a service key, use the `keys generate` command in the Flow CLI.
...
 ```
 
+## Emulating mainnet and testnet transactions
+The emulator allows you to simulate the execution of transactions as if they were 
+performed on the mainnet or testnet. In order to activate this feature, 
+you must specify the network name for the chain ID flag in the following manner:
+```
+flow emulator --chain-id mainnet
+```
+Please note, the actual execution on the real network may differ.
```
janezpodhostnik, line 223:
```suggestion
Please note, the actual execution on the real network may still differ, due to:
- different execution code
- different handling of signatures
- different block formation
```

Might be worth explaining a bit more about why so that the user has a better idea of what they can expect.

janezpodhostnik (approved):
LGTM.

transaction_test.go
janezpodhostnik: we could run the test against mainnet and dump all the registers read into a file, then switch the tests for read from that file.

I guess the file would need to be updated from time to time, but as long as we have some instructions on how to do it it should be fine right?

sideninja: This could work, let me take a stab at it.

Issue #384, opened 2023-05-22
Define a debugger pragma directive


sideninja:
We could define a pragma directive for an emulator to detect and use as a stopping request for the debugger. We could implement this in the emulator to detect and only stop on a transaction that contains this pragma directive. 

Further more we could introduce a way to set a debugger breakpoint using such a pragma, but that would require also changes in the Cadence codebase and should probably be further defined in a Cadence issue.

PR #386, opened 2023-05-22
Prevent stopping debugger on attach

sideninja:
This prevents the debugger from stopping on entry when using the attach mode. The reason for this is that the debugger might stop on an entry of some unrelated execution (system contracts, background noise etc).
When we make the default to false it will only stop on the first transaction the emulator executes. We could further improve this by having a detection of pragma argument that would stop the transaction containing such an argument but I believe this way will be ok for now and we can potentially implement that as part of the issue: https://github.com/onflow/flow-emulator/issues/384

______

For contributor use:


bluesign:
This is not enough I think, also need to remove pause logic in blockchain.go

sideninja:
> This is not enough I think, also need to remove pause logic in blockchain.go ( for example VSCode extension running transaction for account management etc )

That is ok for now I believe because otherwise nothing would stop on entry of a transaction. We can later introduce the pragma directive as discussed that would trigger pause but we need to have at least something to stop.

janezpodhostnik (approved):

PR #391, opened 2023-06-01
Fix breaking changes for `runtime.Config` usage

m-Peter:
## Description

- Remove `CoverageReportedRuntime` and pass `CoverageReport` in `runtime.Config`.
- Remove also the `SetCoverageReport` method and provide instead a function `Option`, namely `WithCoverageReport`.
______

For contributor use:


blockchain.go
```diff
@@ -1620,13 +1613,13 @@ func (b *Blockchain) EndDebugging() {
...
 
 func (b *Blockchain) CoverageReport() *runtime.CoverageReport {
-	return b.coverageReportedRuntime.CoverageReport
+	return b.coverageReport
 }
 
 func (b *Blockchain) SetCoverageReport(coverageReport *runtime.CoverageReport) {
-	b.coverageReportedRuntime.CoverageReport = coverageReport
+	b.coverageReport = coverageReport
 }
```
bluesign, line 1621: I think this is pointless now, as we move with config

bluesign, line 1621: I think this is pointless now, as we move with config

turbolent:
Nice, thank you with starting the update here already.

We'll also need to update to the latest flow-go version, the update is in progress here: https://github.com/onflow/flow-go/pull/4390

m-Peter:
I just opened this as a starting point. It's not done yet, we still need to fix the breaking changes from `flow-go`. I will keep an eye for a new version of `flow-go` and bump the version here. And I certainly need to check it end-to-end, as soon as it builds.

turbolent:
@m-Peter Opened #393 for the update. Could you please rebase your changes on top of it?

blockchain.go
m-Peter: Good catch! I've removed it entirely.

m-Peter:
@turbolent Rebased on master, to include the latest changes, and then on top of the branch you mentioned. Along the way, I did some tech debt refactoring.

PR #429, opened 2023-06-20
Add REST API metrics to /metrics

### LINKED ISSUE #226: Provide Requests Output For Scripts and Transactions
### Issue To Be Solved

Access Nodes have rate limits. For the nodes hosted by Dapper Labs, the limits can be found here:

https://developers.flow.com/nodes/access-api-rate-limits

The rate limits are broken down according to the type of request.

Understanding rate limits are critical to designing and implementing scalable designs/architectures on Flow.

It would be really powerful if developers could switch on the outputting of requests that were made during each transaction or script execution.

This would provide a greater understanding of the requests that are being made to access nodes, allowing developers to make informed decisions when designing their platforms. It would also be a huge help in preventing platforms from hitting rate limits, and provide extra transparency into the inner workings of interactions.

bluesign:
Closes #226 

## Description

Added REST API metrics to /metrics admin endpoint 

______

For contributor use:

### DIFF
--- a/server/access/rest.go
+++ b/server/access/rest.go
@@ -20,12 +20,15 @@ package access
 
 import (
 	"context"
+	"flag"
 	"fmt"
+	"github.com/onflow/flow-go/module"
 	"net"
 	"net/http"
 	"os"
 
 	"github.com/onflow/flow-emulator/adapters"
+	metricsProm "github.com/slok/go-http-metrics/metrics/prometheus"
 
 	"github.com/onflow/flow-go/engine/access/rest"
 	"github.com/onflow/flow-go/model/flow"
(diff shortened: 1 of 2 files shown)


server/access/rest.go
```diff
@@ -78,8 +81,12 @@ func NewRestServer(logger *zerolog.Logger, adapter *adapters.AccessAdapter, chai
 	if debug {
 		debugLogger = zerolog.New(os.Stdout)
 	}
+	var restCollector module.RestMetrics = metrics.NewNoopCollector()
 
-	restCollector := metrics.NewNoopCollector()
+	// only collect metrics if not test
```
turbolent, line 86: When would that be the case? When running the tests of the emulator?

turbolent (approved):
👍

server/access/rest.go
bluesign: yeah emulator tests run in parallel, but prometheus wants single metric instance. So we if we have two REST server instance, it is complaining.

PR #437, opened 2023-06-23
Basic pragma support ( sourceFile and debug )

### LINKED ISSUE #384: Define a debugger pragma directive

### LINKED ISSUE #388: Register source files for scripts/transactions
### Issue To Be Solved

Currently, when viewing coverage reports either from tests or interaction with the emulator, the script/transaction locations are very difficult to interpret, as only the hash ID is presented. For example:

```json
"t.147b164ddb574a408ed4f503df8511caea1887db3bd0444729a0f54ef183b070": {
  "line_hits": {
    "13": 1,
    "14": 0,
    "5": 1,
    "8": 1,
    "9": 1
  },
  "missed_lines": [
    14
  ],
  "statements": 5,
  "percentage": "80.0%"
}
```

Screenshot from 2023-05-22 15-05-00

From this, developers are not able to map a script/transaction hash ID, to its corresponding source file.

### Suggest A Solution

One possible solution would be to keep a global mapping, from script/transaction hash ID, to its corresponding source file. This will be a many-to-one mapping, where many keys (script/transaction IDs) can point to the same source file. This can happen, because the script/transaction hash ID is affected by user-provided arguments, among other things (such as signers for a particular transaction).
This global mapping, can then be used to substitute script/transaction hash IDs, with the mapped source files. We would need to hook into script/transaction execution, and register the corresponding source file.

**Remarks:** The emulator does not deal with source files when executing scripts/transactions. The source code of scripts/transactions is in the form of `[]byte`.

When emulator is used from within `flow-cli` the files are read from the respective commands:
- https://github.com/onflow/flow-cli/blob/master/internal/transactions/send.go#L67
- https://github.com/onflow/flow-cli/blob/master/internal/scripts/execute.go#L61

When emulator is used from within `cadence-tools/test`, the files are read from the testing framework itself (https://github.com/onflow/cadence/blob/master/runtime/stdlib/contracts/crypto_test.cdc#L23).

### Context

This would be useful for tools such as:
- https://github.com/onflow/developer-grants/issues/132
- https://github.com/onflow/developer-grants/issues/178

bluesign:
Closes #384 
Closes #388

## Description

Basic pragma support for `debug` and `sourceFile`

______

For contributor use:

### DIFF
--- a/emulator/blockchain.go
+++ b/emulator/blockchain.go
@@ -64,6 +64,7 @@ func New(opts ...Option) (*Blockchain, error) {
 		activeDebuggingSession: false,
 		conf:                   conf,
 		clock:                  NewSystemClock(),
+		sourceFileMap:          make(map[common.Location]string),
 	}
 	err := b.ReloadBlockchain()
 	if err != nil {
@@ -301,6 +302,8 @@ type Blockchain struct {
 	conf config
 
 	coverageReportedRuntime *CoverageReportedRuntime
+
+	sourceFileMap map[common.Location]string
 }
 
 // config is a set of configuration options for an emulated emulator.
@@ -1174,7 +1177,10 @@ func (b *Blockchain) executeNextTransaction(ctx fvm.Context) (*types.Transaction
 
 	b.currentCode = string(txnBody.Script)
 	b.currentScriptID = txnId.String()
-	if b.activeDebuggingSession {
+
+	pragmas := ExtractPragmas(b.currentCode)
+
+	if b.activeDebuggingSession && pragmas.HasPragma("debug") {
 		b.debugger.RequestPause()
 	}
 
@@ -1194,6 +1200,13 @@ func (b *Blockchain) executeNextTransaction(ctx fvm.Context) (*types.Transaction
 	// if transaction error exist try to further debug what was the problem
 	if tr.Error != nil {
 		tr.Debug = b.debugSignatureError(tr.Error, txnBody)
+	}
+
+	//add to source map if any pragma
+	if pragmas.HasPragma("sourceFile") {
+		location := common.NewTransactionLocation(nil, tr.TransactionID.Bytes())
+		sourceFile := pragmas.ArgumentForPragma("sourceFile")
+		b.sourceFileMap[location] = sourceFile
 	}
 
 	return tr, nil
@@ -1363,7 +1376,10 @@ func (b *Blockchain) executeScriptAtBlockID(script []byte, arguments [][]byte, i
 	scriptProc := fvm.Script(script).WithArguments(arguments...)
 	b.currentCode = string(script)
 	b.currentScriptID = scriptProc.ID.String()
-	if b.activeDebuggingSession {
+
+	pragmas := ExtractPragmas(b.currentCode)
+
+	if b.activeDebuggingSession && pragmas.HasPragma("debug") {
 		b.debugger.RequestPause()
 	}
 	_, output, err := b.vm.Run(
@@ -1388,6 +1404,13 @@ func (b *Blockchain) executeScriptAtBlockID(script []byte, arguments [][]byte, i
 		convertedValue = output.Value
 	} else {
 		scriptError = convert.VMErrorToEmulator(output.Err)
+	}
+
+	//add to source map if any pragma
+	if pragmas.HasPragma("sourceFile") {
+		location := common.NewScriptLocation(nil, scriptID.Bytes())
+		sourceFile := pragmas.ArgumentForPragma("sourceFile")
+		b.sourceFileMap[location] = sourceFile
 	}
 
 	return &types.ScriptResult{
@@ -1554,6 +1577,7 @@ func (b *Blockchain) GetTransactionResultsByBlockID(blockID flowgo.Identifier) (
 	return results, nil
 }
 
+<<<<<<< HEAD
 func (b *Blockchain) GetLogs(identifier flowgo.Identifier) ([]string, error) {
 	txResult, err := b.storage.TransactionResultByID(context.Background(), identifier)
 	if err != nil {
@@ -1569,4 +1593,36 @@ func (b *Blockchain) GetLogs(identifier flowgo.Identifier) ([]string, error) {
 func (b *Blockchain) SetClock(clock Clock) {
 	b.clock = clock
 	b.pendingBlock.SetClock(clock)
+}
+
+func (b *Blockchain) GetSourceFile(location common.Location) string {
+
+	value, exists := b.sourceFileMap[location]
+	if exists {
+		return value
+	}
+
+	addressLocation, isAddressLocation := location.(common.AddressLocation)
+	if isAddressLocation {
+		view := b.pendingBlock.ledgerState.NewChild()
+
+		env := environment.NewScriptEnvironmentFromStorageSnapshot(
+			b.vmCtx.EnvironmentParams,
+			view)
+
+		r := b.vmCtx.Borrow(env)
+		defer b.vmCtx.Return(r)
+
+		code, err := r.GetAccountContractCode(addressLocation)
+
+		if err != nil {
+			return location.ID()
+		}
+		pragmas := ExtractPragmas(string(code))
+		if pragmas.HasPragma("sourceFile") {
+			return pragmas.ArgumentForPragma("sourceFile")
+		}
+	}
+
+	return location.ID()
 }
--- a/emulator/blockchain.go
+++ b/emulator/blockchain.go
@@ -1621,26 +1621,28 @@ func (b *Blockchain) GetSourceFile(location common.Location) string {
 	}
 
 	addressLocation, isAddressLocation := location.(common.AddressLocation)
-	if isAddressLocation {
-		view := b.pendingBlock.ledgerState.NewChild()
+	if !isAddressLocation {
+		return location.ID()
+	}
+	view := b.pendingBlock.ledgerState.NewChild()
 
-		env := environment.NewScriptEnvironmentFromStorageSnapshot(
-			b.vmCtx.EnvironmentParams,
-			view)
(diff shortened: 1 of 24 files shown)


emulator/pragma.go
```diff
@@ -0,0 +1,60 @@
...
+		if p.Name == name {
+			return p.Argument
+		}
+	}
+	return ""
+}
+
+func ExtractPragmas(code string) (result PragmaList) {
+	result = make(PragmaList, 0)
+	r := regexp.MustCompile(`#([a-zA-Z]*)(\(\"([^\"]*?)\"\))?`)
```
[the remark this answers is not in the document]
peterargue: how often will this be called? regex compilations are somewhat expensive, so you can save the overhead by making this a global.

emulator/blockchain.go
```diff
@@ -1180,7 +1183,10 @@ func (b *Blockchain) executeNextTransaction(ctx fvm.Context) (*types.Transaction
 
 	b.currentCode = string(txnBody.Script)
 	b.currentScriptID = txnId.String()
-	if b.activeDebuggingSession {
+
+	pragmas := ExtractPragmas(b.currentCode)
+
+	if b.activeDebuggingSession && pragmas.HasPragma("debug") {
```
peterargue, line 1189: can you move these magic strings to consts in the pragma file? That will also create a convenient place to document them all.

emulator/blockchain.go
```diff
@@ -1641,3 +1664,35 @@ func (b *Blockchain) GetTransactionResultsByBlockID(blockID flowgo.Identifier) (
 	}
 	return results, nil
 }
+
+func (b *Blockchain) GetSourceFile(location common.Location) string {
```
peterargue, line 1668: this isn't called in anywhere in this PR. Does this need to be hooked up to the code coverage system in some way?

peterargue (commented):
I'm not very familiar with how pragmas are generally used, but I mostly focused on the implementation details. Added a few comments. Can you also add some docs describing these and how they are used?

emulator/pragma.go
```diff
@@ -0,0 +1,60 @@
...
+ * limitations under the License.
+ */
+
+package emulator
+
+import (
+	"regexp"
+)
+
+type PragmaList []Pragma
```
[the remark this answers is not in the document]
bluesign: I think we can have same pragma twice later (linters etc), that's why I started it as a list. ( even though methods are working as they are assume unique; `ArgumentForPragma` etc  ) Probably later we can add more functions there.

emulator/pragma.go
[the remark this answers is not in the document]
bluesign: oh I totally made a mistake there, it should be cadence identifier: `[a-zA-Z_]([0-9a-zA-Z_]*)?`, Yeah I agree with making this global too;  I will fix both.

emulator/pragma.go
```diff
@@ -0,0 +1,60 @@
...
+			return p.Argument
+		}
+	}
+	return ""
+}
+
+func ExtractPragmas(code string) (result PragmaList) {
+	result = make(PragmaList, 0)
+	r := regexp.MustCompile(`#([a-zA-Z]*)(\(\"([^\"]*?)\"\))?`)
+	for _, match := range r.FindAllStringSubmatch(code, -1) {
```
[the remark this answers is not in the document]
bluesign: allowlist would be little too limiting; Pragma limit we can for sure put, but not sure how it is on the Cadence side ( if there is any limit ) I think script has a size limit though.

emulator/pragma.go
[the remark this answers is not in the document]
bluesign: We need to limit a bit the scope here I guess: In cadence it is much broader what pragma can allow; such as `#(_x.a ?? a)(_a: "")` is also valid pragma.

I suggest `[a-zA-Z_]([0-9a-zA-Z_]*)?` as identifier, and basic invocation with string.

emulator/blockchain.go
bluesign: this will be separate PR from @m-Peter, to integrate it to code coverage api endpoint. ( it has to also be added to cadence testing etc )

emulator/pragma.go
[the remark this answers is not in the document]
bluesign: I added here: `Contains`, `FilterByName` and `First` to support multiple pragmas with same name

emulator/pragma.go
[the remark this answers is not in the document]
bluesign: Fixed regex to allow `identifier` `[a-zA-Z_]([0-9a-zA-Z_]*)?` invocation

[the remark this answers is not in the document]
bluesign: Moved regex to global

emulator/blockchain.go
m-Peter: Yes indeed, I have already tested this on the code coverage API endpoint. See image:

Screenshot from 2023-06-26 15-22-44

And it works as expected :tada:

emulator/emulator.go
```diff
@@ -20,6 +20,7 @@ package emulator
 
 import (
 	"fmt"
+	"github.com/onflow/cadence/runtime/common"
```
janezpodhostnik, line 23: linting issue.

emulator/pragma_test.go
```diff
@@ -0,0 +1,105 @@
+package emulator_test
```
janezpodhostnik, line 1: missing copyright.

emulator/blockchain.go
```diff
@@ -1641,3 +1664,35 @@ func (b *Blockchain) GetTransactionResultsByBlockID(blockID flowgo.Identifier) (
...
+
+	value, exists := b.sourceFileMap[location]
+	if exists {
+		return value
+	}
+
+	addressLocation, isAddressLocation := location.(common.AddressLocation)
+	if isAddressLocation {
```
janezpodhostnik, line 1676: Maybe invert if condition to reduce the amount of nested code.

emulator/pragma.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+var (
+	PragmaDebug      = "debug"
+	PragmaSourceFile = "sourceFile"
+)
+
+type PragmaList []Pragma
+
+var pragmaRegexp = regexp.MustCompile(`#([a-zA-Z_]([0-9a-zA-Z_]*)?)(\(\"([^\"]*?)\"\))?`)
```
janezpodhostnik, line 32: can you add a comment with a few examples? Regexes are hard to read.

janezpodhostnik (commented):
Nice! Just left some minor comments.

emulator/pragma.go
bluesign: added in https://github.com/onflow/flow-emulator/pull/437/commits/e6360c624328bcfc657ec1f4fbce5acd8e2fbb86

emulator/blockchain.go
bluesign: inverted in https://github.com/onflow/flow-emulator/pull/437/commits/5f8fbd3b74599ad16d972bfe25efac4c7854462f

emulator/pragma.go
```diff
@@ -0,0 +1,110 @@
...
+		if p.Name() == name {
+			c = c + 1
+		}
+	}
+	return c
+}
+
+func ExtractPragmas(code string) (result PragmaList) {
+	result = make(PragmaList, 0)
+	for _, match := range pragmaRegexp.FindAllStringSubmatch(code, -1) {
```
turbolent, line 103: Instead of using a regular expression it might be better to use the parser and find pragmas in the resulting AST.

Using a regular expression might give false-positives, like if the pragma is commented out

bluesign: hmm this is very good idea. Let me try that quick

bluesign: fixed in https://github.com/onflow/flow-emulator/pull/437/commits/80401416bc2e208d2240d8c29949cd5e7147581e

@turbolent can you have a look if it is ok ?

turbolent: Nice!

emulator/pragma.go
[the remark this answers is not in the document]
bluesign: switched to cadence to parse pragmas, no more regex

emulator/pragma.go
```diff
@@ -0,0 +1,128 @@
...
+				}
+				argument = stringParameter.Value
+			}
+
+			result = append(result, &BasicPragma{
+				name:     expression.InvokedExpression.String(),
+				argument: argument,
+			})
+
+		}
```
turbolent, line 124: The element type check is unnecessary here, given the next line does an element type check too:

```suggestion
		pragmaDeclaration, ok := el.(*ast.PragmaDeclaration)
		if !ok {
			return
		}
		expression, ok := pragmaDeclaration.Expression.(*ast.InvocationExpression)
		if !ok {
			return
		}

		if len(expression.Arguments) > 1 {
			return
		}
		
		var argument string
		if len(expression.Arguments) == 1 {
			stringParameter, ok := expression.Arguments[0].Expression.(*ast.StringExpression)
			if !ok {
				return
			}
			argument = stringParameter.Value
		}

		result = append(result, &BasicPragma{
			name:     expression.InvokedExpression.String(),
			argument: argument,
		})
```

janezpodhostnik (approved):

sideninja (approved):
Awesome feature.

sideninja:
@bluesign can we merge?

bluesign:
yep I rebased yesterday, should be good to go @sideninja

PR #449, opened 2023-07-14
Introduce a settable `Clock` for `Blockchain` and `PendingBlock`

m-Peter:
## Description

Currently, [to measure time in Cadence](https://developers.flow.com/cadence/measuring-time), developers can use something like the following:

```cadence
getCurrentBlock().timestamp
```

Contracts can use the above value in certain conditions, such as unlocking tokens after a year.

To simulate/test scenarios with time expiration or generally passage of time, it would be useful to have the ability to set the blockchain's timestamp in the future or even the past. For example, this can be used and exposed in the Cadence Testing Framework, and in other tools that use the Emulator for integration tests.
 
______

For contributor use:

### DIFF
--- a/emulator/blockchain.go
+++ b/emulator/blockchain.go
@@ -61,6 +61,7 @@ func New(opts ...Option) (*Blockchain, error) {
 		debugger:               nil,
 		activeDebuggingSession: false,
 		conf:                   conf,
+		clock:                  NewSystemClock(),
 	}
 	err := b.ReloadBlockchain()
 	if err != nil {
@@ -280,6 +281,7 @@ type Blockchain struct {
 
 	// pending block containing block info, register state, pending transactions
 	pendingBlock *pendingBlock
+	clock        Clock
 
 	// used to execute transactions and scripts
 	vm    *fvm.VirtualMachine
@@ -399,7 +401,7 @@ func (b *Blockchain) ReloadBlockchain() error {
 		return err
 	}
 
-	b.pendingBlock = newPendingBlock(latestBlock, latestLedger)
+	b.pendingBlock = newPendingBlock(latestBlock, latestLedger, b.clock)
 	b.transactionValidator = configureTransactionValidator(b.conf, blocks)
 
 	return nil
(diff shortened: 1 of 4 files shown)


emulator/blockchain.go
```diff
@@ -1641,3 +1643,8 @@ func (b *Blockchain) GetTransactionResultsByBlockID(blockID flowgo.Identifier) (
 	}
 	return results, nil
 }
+
+func (b *Blockchain) SetTimestamp(timestamp int64) {
+	b.timestamp = timestamp
+	b.pendingBlock.SetTimestamp(b.timestamp)
+}
```
bluesign, line 1650: Maybe we can do this something like `SetTimeDelta`. Most use cases will be about `moving time` ( offer expires after 1 day, valid until 1 year etc )

m-Peter: That's a valid point :+1: I just opened the PR to get some brainstorming on what's the best solution to tackle this problem.

m-Peter: @bluesign Do you have any API blueprint for your suggestion? To use as a point of reference.

bluesign: I think this is good for this PR, we can later advance on this. ( maybe add little mocking capabilities etc )

m-Peter: As a developer, would you prefer the `SetTimestamp` to have effect only for the next block? Or it should be persistent until the developer decides to move the back in the present? With something like `Time.now()`

janezpodhostnik:
I think it might be better to change the `blockchain.go` to have a settable clock:

```go
type Clock interface {
    Now() time.Time
}
```

and `newPendingBlock` would either get a `time.Time` from the clock or a `Clock` as a parameter.

What do you think?

m-Peter:
@janezpodhostnik that's quite a great suggestion 💯 since emulator is also a library, that is a more solid way to enable the injection of any clock implementation. I will make the changes and ping you

m-Peter:
@janezpodhostnik I have addressed your suggestion in this commit: https://github.com/onflow/flow-emulator/pull/449/commits/381235b95366e8ab35e02b83c048e63801da65b2. Let me know if it is close to your original idea :pray:

janezpodhostnik (approved):
This looks a lot better

turbolent (commented):
Nice!

PR #459, opened 2023-08-04
Fix Cadence's unsafe random

tarakby:
Closes https://github.com/onflow/cadence/issues/2691

Emulator now supports Cadence random function(s). The randoms are currently deterministic and equal for all blocks.

______

For contributor use:

### DIFF
--- a/emulator/blockchain.go
+++ b/emulator/blockchain.go
@@ -530,6 +530,19 @@ func (h CadenceHook) Run(_ *zerolog.Event, level zerolog.Level, msg string) {
 	}
 }
 
+// `dummyEntropyProvider“ implements `environment.EntropyProvider`
+// which provides a source of entropy to fvm context (required for Cadence's randomness).
+type dummyEntropyProvider struct{}
+
+var dummySource = make([]byte, 32)
+
+func (gen *dummyEntropyProvider) RandomSource() ([]byte, error) {
+	return dummySource, nil
+}
+
+// make sure `dummyEntropyProvider“ implements `environment.EntropyProvider`
+var _ environment.EntropyProvider = (*dummyEntropyProvider)(nil)
+
 func configureFVM(blockchain *Blockchain, conf config, blocks *blocks) (*fvm.VirtualMachine, fvm.Context, error) {
 	vm := fvm.NewVirtualMachine()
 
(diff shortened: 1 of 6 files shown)


emulator/blockchain.go
```diff
@@ -530,6 +530,19 @@ func (h CadenceHook) Run(_ *zerolog.Event, level zerolog.Level, msg string) {
 	}
 }
 
+// `dummyEntropyProvider“ implements `environment.EntropyProvider`
+// which provides a source of entropy to fvm context (required for Cadence's randomness).
+type dummyEntropyProvider struct{}
+
+var dummySource = make([]byte, 32)
```
tarakby, line 537: This causes randomness on emulator to be:
 1. deterministic
 2. same for all txs/blocks
 
 Feel free to update this to be any behaviour you prefer.

janezpodhostnik:
> does this break mainnet/testnet transaction replay ( simulation ) ? Is there a way we can solve this ? maybe by getting block (transaction ) random seed from archive nodes or similar?

yes, it does break it... but, like you said, we can fix it by exposing the random beacon. 
@tarakby is there anyway to access the random beacon for sealed blocks?

bluesign:
> > does this break mainnet/testnet transaction replay ( simulation ) ? Is there a way we can solve this ? maybe by getting block (transaction ) random seed from archive nodes or similar?
> 
> yes, it does break it... but, like you said, we can fix it by exposing the random beacon. @tarakby is there anyway to access the random beacon for sealed blocks?

maybe we can add to transaction result as random seed.

tarakby:
> @tarakby is there anyway to access the random beacon for sealed blocks?

The issue with the source of randomness is that it is part of the block payload data not the execution state. Archive nodes and other nodes in general are not required to store these data for all the history (most of them only store data since the root  block of the current spork only, but not before that).

> maybe we can add to transaction result as random seed.

Yes this is a good idea - We can restrict it only to transactions that used Cadence's random function, since most of the transactions do not need it. 

I have no idea how to plug that seed into the emulator later when replaying a transaction. Does this PR need further changes to accommodate that?

bluesign (approved):

bluesign:
I think we can merge this in the mean time, for sure random seed in tx result would take some time

btspoony:
Hi @turbolent, Approximately when can this PR be included in a certain release of `flow-cli`?

turbolent:
@btspoony I only assist with reviews in this project. Please open an issue in this and the CLI repo. Thank you!

PR #433, opened 2023-06-22
Rollback and Snapshot support for remote storage

bluesign:
## Description

Added rollback and snapshot support for remote storage 
Remote storage now can use any storage backend ( redid etc ) 
Added some sanity checks ( don't query blocks on archive after fork etc ) 
______

For contributor use:

- [X] Targeted PR against `master` branch
- [ ] Linked to GitHub issue with discussion and accepted design OR link to spec that describes this work
- [X] Code follows the [standards mentioned here](https://github.com/onflow/flow-emulator/blob/master/CONTRIBUTING.md#styleguides)
- [ ] Updated relevant documentation
- [X] Re-reviewed `Files changed` in the GitHub PR explorer
- [ ] Added appropriate labels

PR #489, opened 2023-10-25
Improve block handling in remote store

peterargue:
Closes #???

## Description

The forked mainnet functionality is helpful to testing full application integration. With the current implementation, there are some key differences between behavior of the real network and emulator, particularly with error messages and block handling.

This PR attempts to close the gap and adds testing to cover various cases to ensure block responses behave as expected.

______

For contributor use:


Makefile
```diff
@@ -15,7 +15,7 @@ GOPATH ?= $(HOME)/go
 install-tools:
 	mkdir -p ${GOPATH}; \
 	cd ${GOPATH}; \
-	GO111MODULE=on go install github.com/golang/mock/mockgen@v1.6.0; \
+	GO111MODULE=on go install go.uber.org/mock/mockgen@v0.3.0; \
```
peterargue, line 18: `github.com/golang/mock` is deprecated

storage/remote/store.go
```diff
@@ -176,11 +224,6 @@ func (s *Store) LedgerByHeight(
 	ctx context.Context,
 	blockHeight uint64,
 ) (snapshot.StorageSnapshot, error) {
-	err := s.SetBlockHeight(blockHeight)
```
peterargue, line 179: Setting this here could have confusing results, since it may reset the latest height to an earlier block if `ExecuteScript` is called with a past height.

It's already called when committing a block, so we can skip it
https://github.com/onflow/flow-emulator/blob/0273f8fe3807746ac065b64dce30c8deafd7c216/storage/store.go#L213-L216

peterargue: @sideninja keep me honest

bluesign:
@peterargue thanks a lot for this, btw do we have archive nodes still ? last time I tried to work on this I was failing at that. 

Also if someone can fix, or at least explain to me, how flow-archive tags works, it would be great, there are 2 version stream one with v1.x another one is v0.x but I guess newer is v0.x but last time I tried archive node was using v1.x etc

sideninja:
> @peterargue thanks a lot for this, btw do we have archive nodes still ? last time I tried to work on this I was failing at that.
> 
> Also if someone can fix, or at least explain to me, how flow-archive tags works, it would be great, there are 2 version stream one with v1.x another one is v0.x but I guess newer is v0.x but last time I tried archive node was using v1.x etc

We are working on merging archive node functionality into access nodes as we speak. So right now the issues you are experiencing are due to that. We should have something available soon.

storage/remote/store.go
sideninja: Yes, looking at this it doesn't need to be here, not sure what was my thinking at the time.

sideninja (approved):

storage/remote/store_test.go
```diff
@@ -287,3 +292,224 @@ func Test_SimulatedMainnetTransactionWithChanges(t *testing.T) {
 	assert.Equal(t, txRes.Events[0].String(), "A.9799f28ff0453528.Ping.PingEmitted: 0x953f6f26d61710cb0e140bfde1022483b9ef410ddd181bac287d9968c84f4778")
 	assert.Equal(t, txRes.Events[0].Value.String(), `A.9799f28ff0453528.Ping.PingEmitted(sound: "pong pong pong")`)
 }
+
+func Test_SimulatedMainnetProgressesBlocks(t *testing.T) {
```
sideninja, line 296: nice!

storage/remote/store.go
```diff
@@ -176,11 +224,6 @@ func (s *Store) LedgerByHeight(
...
-	if err != nil {
-		return nil, err
-	}
-
 	return snapshot.NewReadFuncStorageSnapshot(func(id flowgo.RegisterID) (flowgo.RegisterValue, error) {
 		// first try to see if we have local stored ledger
 		value, err := s.DefaultStore.GetBytesAtVersion(
```
bluesign, line 229: I think here needs a check, something like: 

```go
if blockHeight <= s.CurrentHeight{
    value, err := s.DefaultStore.GetBytesAtVersion(
        ctx,
        s.KeyGenerator.Storage(storage.LedgerStoreName),
        []byte(id.String()),
        blockHeight,
	)
}

```

storage/remote/store.go
```diff
@@ -193,6 +236,7 @@ func (s *Store) LedgerByHeight(
 			return value, nil
 		}
 
+		log.Printf("fetching register: owner: %x, key: %x", []byte(id.Owner), []byte(id.Key))
```
turbolent, line 239: Is this development/debug code? Maybe remove if not needed

turbolent (commented):
Nice!

storage/remote/store.go
peterargue: great point. I think this should just rewrite the height if it's greater than the fork height

storage/remote/store.go
peterargue: oops. good catch

bluesign:
I think we need to fix also register requests part, I messed up there before because of flow-archive versioning problems. ( I can make a PR for it, or contribute to this PR )

bluesign:
ok I updated https://github.com/onflow/flow-emulator/pull/433 with required changes, which required a PR on flow-archive ( for conflicting go versions ) https://github.com/onflow/flow-archive/pull/143

PR #491, opened 2023-10-31
Fix docs on snapshot persistence and add tests around persistence

### LINKED ISSUE #490: Snapshot flag (`--snapshot`) doesn't automatically enable state persistance
### Problem

I've noticed that when I run `flow emulator --snapshot`, created snapshots aren't persisted through emulator restarts, contrary to what [the documentation says](https://github.com/onflow/flow-emulator#configuration):

Screenshot 2023-10-29 at 15 28 09

### Steps to Reproduce

1. run `flow emulator --snapshot`
2. create a snapshot
3. restart the emulator
4. list snapshots endpoint returns null (http://localhost:8080/emulator/snapshots)

But when you run `flow emulator --snapshot --persist`, snapshots are persisted as expected.

### Acceptance Criteria

Snapshots are automatically persisted when `--snapshot` flag is used, as there is no use-case in using snapshots without persistence.

### Context

Working on implementing automatic project state saving (using snapshots) in Flowser.

chasefleming:
Closes #490 

______

For contributor use:


server/server.go
```diff
@@ -327,7 +327,7 @@ func configureStorage(conf *Config) (storageProvider storage.Store, err error) {
 		}
 	}
 
-	if conf.Persist {
+	if conf.Persist || conf.Snapshot {
```
nvdtf, line 330: do we need to update `conf.Persist` to `true` so if subsequent code relies on the flag it would work as expected?

bluesign:
Sorry @chasefleming just saw the issue, actually documentation needs update, implementation is correct. 

We have for some time possibility of snapshots without persistence, unfortunately when I made the storage reorg PR, I forgot the documentation update probably.

chasefleming:
No worries @bluesign was fun to dig into the code. Do you want me to remove the change but keep some of the tests so that there are more tests? I can update the documentation as well in this to be correct. Let me know.

chasefleming:
@bluesign I fixed the documentation and updated the tests to reflect what you said

bluesign (approved):

PR #496, opened 2023-11-07
REST API streaming support 

### LINKED ISSUE #495: Support state stream API
### Problem

The emulator does not currently implement the state stream API needed for events streaming

### Steps to Reproduce

Try using events streaming (i.e. `ws://localhost:8888/v1/subscribe_events`) and the endpoint will not be available because the state_stream.API instance used to instantiate the REST server is nil https://github.com/onflow/flow-emulator/blob/0cdc31ceecde40ee7ffbf70b8e77961b7bbd7ce4/server/access/rest.go#L99-L117

### Acceptance Criteria

Event streaming supported by emulator

### Context

Blocking adoption of event streaming API (i.e. https://github.com/onflow/fcl-js/issues/1791)

bluesign:
Closes #495 

## Description

Currently trie updates are not implemented but as-is it is enough to provide events, currently REST API does not allow state streaming AFAIK, so I marked as ready for review.  ( as it is a blocker for FCL work ) 

______

For contributor use:


bluesign:
@peterargue can you have a look at this? I am not sure if we need execution streaming here or just events are enough

server/access/streamBackend.go
```diff
@@ -0,0 +1,221 @@
...
+		getStartHeight:   getStartHeightFunc(blockchain),
+	}
+}
+
+var _ state_stream.API = &StateStreamBackend{}
+
+func getStartHeightFunc(blockchain *emulator.Blockchain) GetStartHeightFunc {
+	return func(blockID flow.Identifier, height uint64) (uint64, error) {
+		block, err := blockchain.GetBlockByID(blockID)
+		if err == nil {
```
sideninja, line 70: Maybe worth checking type of an error: 
```go
errors.Is(err, storage.ErrNotFound)
```

server/access/streamBackend.go
```diff
@@ -0,0 +1,221 @@
...
+		}
+		return height, nil
+	}
+}
+
+func getExecutionDataFunc(blockchain *emulator.Blockchain) GetExecutionDataFunc {
+	return func(_ context.Context, height uint64) (*execution_data.BlockExecutionDataEntity, error) {
+		block, err := blockchain.GetBlockByHeight(height)
+
+		if err != nil {
```
sideninja, line 81: what if the returned error is not storage err not found? should you overwrite an error like so?

server/access/streamBackend.go
```diff
@@ -0,0 +1,221 @@
...
+			return nil, storage.ErrNotFound
+		}
+		chunks := make([]*execution_data.ChunkExecutionData, 0)
+
+		for _, collectionGuarantee := range block.Payload.Guarantees {
```
sideninja, line 86: nit: move after the `blockchain.GetCollectionByID`

server/access/streamBackend.go
```diff
@@ -0,0 +1,221 @@
...
+				if err != nil {
+					return nil, err
+				}
+				events = append(events, txResult.Events...)
+
+				var lightResult flow.LightTransactionResult
+				lightResult.TransactionID = txResult.TransactionID
+				lightResult.ComputationUsed = 0
+				lightResult.Failed = txResult.Status != 0
+				txResults = append(txResults, lightResult)
```
sideninja, line 112: nit change to inline:
```go
lightResult := flow.LightTransactionResult{
    TransactionID: txResult.TransactionID,
    Failed: txResult.ErrorMessage != "",
}
```
not so nit:
Failed should be determined by checking the presence of an error: `Failed: txResult.ErrorMessage != ""`.

server/access/streamBackend.go
```diff
@@ -0,0 +1,221 @@
...
+			}
+
+			chunk := &execution_data.ChunkExecutionData{
+				Collection: collection,
+				Events:     events,
+				//TODO: add trie updates
+				TrieUpdate:         &ledger.TrieUpdate{},
```
sideninja, line 119: nit: remove this line, it's better to default to nil.

server/access/streamBackend.go
```diff
@@ -0,0 +1,221 @@
...
+				TransactionResults: txResults,
+			}
+			chunks = append(chunks, chunk)
+		}
+
+		result := execution_data.NewBlockExecutionDataEntity(
+			flow.ZeroID,
```
sideninja, line 126: I believe you can calculate the ID like this:
```go
buf := new(bytes.Buffer)

serializer := execution_data.DefaultSerializer
err = serializer.Serialize(buf, executionData)
if err != nil {
    return nil, err
}

rootBlob := blobs.NewBlob(buf.Bytes())
rootID, err := flow.CidToId(rootBlob.Cid())
if err != nil {
    return nil, err
}

result := execution_data.NewBlockExecutionDataEntity(
    rootID,
    executionData,
)
```

server/access/streamBackend.go
bluesign: is the logic here to try with blockID first then try with block height?

server/access/streamBackend.go
bluesign: https://github.com/onflow/flow-emulator/pull/496/commits/0361e71aaf83d2e7c89c171301647cc8e4e775f8

server/access/streamBackend.go
bluesign: https://github.com/onflow/flow-emulator/pull/496/commits/51b9ff62aa912fdb5288b65926db30c9266b592e

server/access/streamBackend.go
bluesign: https://github.com/onflow/flow-emulator/pull/496/commits/46deb1901110a133cc1a5475b92ebf89706e2898

server/access/streamBackend.go
bluesign: I didn't quite get this, you mean the  allocation ?

server/access/streamBackend.go
bluesign: handled the not found case separately

server/access/streamBackend.go
sideninja: I would say the logic is fine, but if GetBlockByID is not found then return the height, but if the GetBlockByID returns an unknown error then I believe we should return that error.

server/access/streamBackend.go
sideninja: I mean the line 87 could be moved to line 92 after the call... super nit so ignore :D

sideninja (approved):

server/access/streamBackend.go
bluesign: done :)

server/access/streamBackend.go
bluesign: @sideninja this failed to provide ID unfortunately :(

sideninja:
> > @bluesign good to merge?
> 
> Should be good, just this ID I will investigate a bit later but we can merge

yeah doesn't matter as much for now

Issue #509, opened 2023-11-16
Improve Randomness in Emulator


chasefleming:
## Current Behavior

In the Flow emulator, the randomness is derived from a constant source of randomness. This constant usage results in repetitive randomness patterns.

### Ideal Behavior:

- **Dynamic Randomness per Block:** For a better user experience, particularly when simulating transactions/scripts against different blocks, it would be ideal to have different random values per block.
- **Utilizing Block ID:** Instead of the hardcoded SoR constant, the suggestion is to use a block ID (if present) in the emulator to generate the SoR. This would create dynamic randomness per block.

## Impact and Rationale

- **Enhances Testing Realism:** This change would make emulator testing more reflective of real-world scenarios, where randomness varies per block.
- **No Security Concerns for Emulator:** Using a block hash is not secure for TestNet/MainNet, but in the emulator context, security is not a primary concern, making this approach fine for improving user experience.

bluesign:
note: here we should also consider debugging mainnet/testnet transactions, probably we will not have access to SoR but we can at least mention it in the docs maybe.

PR #511, opened 2023-11-16
Bring randomness in Emulator on par with `testnet` & `mainnet`

m-Peter:
Closes https://github.com/onflow/flow-emulator/issues/509

## Description

- Configure `FVM` with `fvm.WithRandomSourceHistoryCallAllowed(true)` option
- Introduce a system chunk transaction, for calling the `RandomBeaconHistory.heartbeat` method
- Add a second key on service account, to be able to sign the system chunk transaction, without conflicts on sequence number
- Improve `dummyEntropyProvider` to use the latest block ID for deriving the source of entropy

______

For contributor use:

### DIFF
--- a/emulator/accounts_test.go
+++ b/emulator/accounts_test.go
@@ -836,7 +836,7 @@ func TestRemoveAccountKey(t *testing.T) {
 	account, err = adapter.GetAccount(context.Background(), b.ServiceKey().Address)
 	assert.NoError(t, err)
 
-	// key at index 1 should not be revoked
+	// keys at indices 1 and 2 should not be revoked
 	require.Len(t, account.Keys, 3)
 	assert.True(t, account.Keys[0].Revoked)
 	assert.False(t, account.Keys[1].Revoked)
--- a/emulator/accounts_test.go
+++ b/emulator/accounts_test.go
@@ -677,7 +677,7 @@ func TestAddAccountKey(t *testing.T) {
 
 		script := []byte("transaction { execute {} }")
 
-		var newKeyID = 2 // new key with have ID 1
+		var newKeyID = 1 // new key will have ID 1
 		var newKeySequenceNum uint64 = 0
 
 		tx2 := flowsdk.NewTransaction().
@@ -768,10 +768,9 @@ func TestRemoveAccountKey(t *testing.T) {
 	account, err := adapter.GetAccount(context.Background(), b.ServiceKey().Address)
 	assert.NoError(t, err)
 
-	require.Len(t, account.Keys, 3)
+	require.Len(t, account.Keys, 2)
 	assert.False(t, account.Keys[0].Revoked)
 	assert.False(t, account.Keys[1].Revoked)
-	assert.False(t, account.Keys[2].Revoked)
 
 	// create transaction that removes service key
 	tx2 := templates.RemoveAccountKey(b.ServiceKey().Address, 0)
@@ -801,13 +800,12 @@ func TestRemoveAccountKey(t *testing.T) {
 	assert.NoError(t, err)
 
 	// key at index 0 should be revoked
-	require.Len(t, account.Keys, 3)
+	require.Len(t, account.Keys, 2)
 	assert.True(t, account.Keys[0].Revoked)
 	assert.False(t, account.Keys[1].Revoked)
-	assert.False(t, account.Keys[2].Revoked)
 
-	// create transaction that removes the last account key
-	tx3 := templates.RemoveAccountKey(b.ServiceKey().Address, 2)
+	// create transaction that removes remaining account key
+	tx3 := templates.RemoveAccountKey(b.ServiceKey().Address, 0)
 
 	tx3.SetGasLimit(flowgo.DefaultMaxTransactionGasLimit).
 		SetProposalKey(b.ServiceKey().Address, b.ServiceKey().Index, b.ServiceKey().SequenceNumber).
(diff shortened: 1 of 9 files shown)


m-Peter:
**Note:** Tests for the new additions are pending :pray:

emulator/accounts_test.go
```diff
@@ -835,19 +837,20 @@ func TestRemoveAccountKey(t *testing.T) {
 	assert.NoError(t, err)
 
 	// key at index 1 should not be revoked
```
janezpodhostnik, line 839:
```suggestion
	// key at indices 1 and 2 should not be revoked
```

emulator/blockchain.go
```diff
@@ -585,7 +598,8 @@ func configureFVM(blockchain *Blockchain, conf config, blocks *blocks) (*fvm.Vir
 		fvm.WithAccountStorageLimit(conf.StorageLimitEnabled),
 		fvm.WithTransactionFeesEnabled(conf.TransactionFeesEnabled),
 		fvm.WithReusableCadenceRuntimePool(customRuntimePool),
-		fvm.WithEntropyProvider(&dummyEntropyProvider{}),
+		fvm.WithEntropyProvider(blockchain.entropyProvider),
+		fvm.WithRandomSourceHistoryCallAllowed(true),
```
janezpodhostnik, line 602: Ideally this would only be enabled for the system transaction.

emulator/blockchain.go
```diff
@@ -1182,6 +1208,34 @@ func (b *Blockchain) executeBlock() ([]*types.TransactionResult, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	err = tx.SignEnvelope(serviceKey.Address, 1, signer)
+	if err != nil {
+		return nil, err
+	}
+	txn := convert.SDKTransactionToFlow(*tx)
+	b.pendingBlock.AddTransaction(*txn)
```
janezpodhostnik, line 1237: I would take a note of which transaction index this is. then later in this function when executing the transaction with this index do `ctx :=  fvm.NewContextFromParent(blockContext,  fvm.WithRandomSourceHistoryCallAllowed(true))`

janezpodhostnik (commented):
Very nice 👏

tarakby (approved):
👌🏼

emulator/blockchain.go
```diff
@@ -1182,6 +1208,34 @@ func (b *Blockchain) executeBlock() ([]*types.TransactionResult, error) {
...
+	signer, err := serviceKey.Signer()
+	if err != nil {
+		return nil, err
+	}
+
+	err = tx.SignEnvelope(serviceKey.Address, 1, signer)
+	if err != nil {
+		return nil, err
+	}
+	txn := convert.SDKTransactionToFlow(*tx)
```
bluesign, line 1236: I am not sure about this part, @janezpodhostnik knows better, maybe we have another way to execute this transaction without modifying SeqNumber of the service account, this may break some tests maybe? if they were assuming seqNumber increasing if they send transaction by one

emulator/accounts_test.go
m-Peter: Updated in https://github.com/onflow/flow-emulator/pull/511/commits/2ee07522ad27fc650a07df3b7748b769493ad517

emulator/blockchain.go
m-Peter: That's a good catch :100:

emulator/blockchain.go
m-Peter: Good catch :100: 
I forgot to enable this just for the system chunk transaction.
Updated in https://github.com/onflow/flow-emulator/pull/511/commits/2ee07522ad27fc650a07df3b7748b769493ad517
I was able to make do without using a transaction index for this :pray:

emulator/blockchain.go
m-Peter: @bluesign  I tried to use the predefined service key, as the proposal key, but that wouldn't work. Even a transaction for account creation is signed by the service account, which means that in the same block, we can't have two transactions from the service account, with the same proposal key. That's why I added a completely new account key. This in fact did break some tests, but I've fixed them. Not sure if there's an alternative for this, because I would always get a conflict on the sequence number, when using the same proposal key.

bluesign: I think one option is to send proposer at `0x0` (actually not setting proposer at all is same effect I guess) and temporarily disable transaction validation. I think somehow real network is not setting proposer and signature for this transaction too

m-Peter: Aha, I see. I wasn't able to find how this transaction body is constructed in the `flow-go` repository.

bluesign: I tried to make a PR to your repo but didn't allow me @m-Peter 

Basically what I did was : 

- move chunkTransaction execution to `commitBlock` ( as we can skip blocks with commitBlock without executing)
- rollback test changes with results ( length of results ) 
- execute chunkTransaction with:
```go
ctx := fvm.NewContextFromParent(
		b.vmCtx,
		fvm.WithAuthorizationChecksEnabled(false),
		fvm.WithSequenceNumberCheckAndIncrementEnabled(false),
		fvm.WithRandomSourceHistoryCallAllowed(true),
	)
```

m-Peter: That's nice :clap: 
So the checks are enabled/disabled on the context itself. I thought it was on transaction body :sweat_smile: 
Not sure why you can't open a PR, it is just a fork of this one.
Let me see if I can apply your suggestions :100:

bluesign: I am not sure about events part: `b.pendingBlock.events = append(b.pendingBlock.events, output.Events...)`
not sure how mainnet manage systemchunkevents, @janezpodhostnik can chime in there.

m-Peter: I do not think that the `RandomBeaconHistory` contract emits any events. Maybe I misunderstood something though?

bluesign: oh I remember there was a discussion on events there, if there is none there is no concern.

m-Peter: Hats off to @bluesign for his suggestion :100: 
That clearly made the changes more contained, as there are less files affected.
Take a look and let me know if this is what you had in mind.

emulator/blockchain.go
```diff
@@ -1335,6 +1348,35 @@ func (b *Blockchain) commitBlock() (*flowgo.Block, error) {
...
+	if err != nil {
+		panic(err)
+	}
+
+	b.pendingBlock.events = append(b.pendingBlock.events, output.Events...)
+
+	err = b.pendingBlock.ledgerState.Merge(executionSnapshot)
+	if err != nil {
+		panic(err)
+	}
```
sideninja, line 1379: extract this into a function, something like updateRandomSourceHistory

emulator/blockchain.go
```diff
@@ -1699,3 +1741,21 @@ func (b *Blockchain) GetSourceFile(location common.Location) string {
...
+	script := templates.ReplaceAddresses(
+		systemChunkTransactionTemplate,
+		templates.Environment{
+			RandomBeaconHistoryAddress: b.GetChain().ServiceAddress().Hex(),
+		},
+	)
+
+	tx := flowgo.NewTransactionBody().
+		SetScript([]byte(script)).
+		SetGasLimit(9999).
```
sideninja, line 1755: use the flow-go constant for max gas

sideninja (approved):

emulator/blockchain.go
bluesign: great work @m-Peter

bluesign (approved):
great work

emulator/blockchain.go
m-Peter: Good point :+1: 
Updated in https://github.com/onflow/flow-emulator/pull/511/commits/e1686844a60e3165344b08bb9be89fa695a5bf26

emulator/blockchain.go
m-Peter: Good point :+1: 
Updated in https://github.com/onflow/flow-emulator/pull/511/commits/e1686844a60e3165344b08bb9be89fa695a5bf26

Issue #534, opened 2023-12-14
Docker builds are broken


turbolent:
### Problem

CI automatically builds Docker images. They're currently broken, see e.g. https://github.com/onflow/flow-emulator/actions/runs/7212048234/job/19649026562#step:7:648

The build errors show problems in the crypto library.

### Steps to Reproduce

See link above / "Docker Image" CI job's logs.

### Acceptance Criteria

Docker images are built correctly.

tarakby:
@sideninja and I faced these problems yesterday with the CLI build as well. 
First we need to enable cgo with `CGO_ENABLED=1`, it's also preferred to add the flag `CGO_CFLAGS="-O -D__BLST_PORTABLE__"` in case the building or target machines are old (this should be [added](https://github.com/onflow/flow-emulator/blob/473f20fc67a907bb8b001bb5ed800a30f7db2956/cmd/emulator/Dockerfile#L24) to the `Dockerfile`), but then the main problem becomes cross-compilation. I can see the repo [targets both](https://github.com/onflow/flow-emulator/blob/473f20fc67a907bb8b001bb5ed800a30f7db2956/.github/workflows/ci.yml#L106) x86 and aarch64 .
Not sure if one of the two builds had succeeded from the logs. I guess it also depends on the architecture of the github action machine. But anyway the `Dockerfile` needs to be updated to handle cross-compilation properly, since CGO is now being used. 

I was planning to work on these tasks with the new crypto library repo by repo, but things went fast with the EVM releases. 
A quick temp fix that worked for CLI is to go back to the previous crypto library `onflow/flow-go/crypto@v.0.24.9` in a replace statement.

turbolent:
This also breaks building the Cadence language server to WebAssembly. Before it was possible to build without CGo, how come we need it now?

tarakby:
> This also breaks building the Cadence language server to WebAssembly 

Does the Cadence language server depend on the emulator?

> how come we need it now?

In the past there was a build tag that is needed if BLS stuff need to be compiled. CLI and emulator always avoided that build tag and excluded BLS (mainly to avoid an extra setup step needed for BLS, not to avoid cgo). The new crypto module doesn't require the setup build, so the build tag was removed.

Can you please point me to the build script of the Cadence language server?

bluesign:
does emulator need BLS? I don't know if evm stuff needs it, but if not maybe we can use build tag again.

tarakby:
[...]

PR #548, opened 2024-01-10
Specify C compiler in build cross-compilation

tarakby:
Closes https://github.com/onflow/flow-emulator/issues/534

## Description

- remove unused `Dockerfile.relic`
- update `Dockerfile` go build step:
  - enable cgo. 
  - specify the C compiler for each arch target. This insures cross-compilation of the crypto non-portable code succeeds.
  - disable ADX instructions when compiling the crypto lib: crypto operations become slower but it makes sure binary can run on all CPUs.
  
Builds were enabled temporarily while working on the PR. You can see the results [here](https://github.com/onflow/flow-emulator/actions/runs/7506865261/job/20439222271).
______

For contributor use:


cmd/emulator/Dockerfile
```diff
@@ -12,16 +12,28 @@ WORKDIR /app
...
 RUN git config --global 'url.ssh://git@github.com/.insteadOf' https://github.com/
+RUN apt-get update
+RUN apt-get -y install apt-utils gcc-arm-linux-gnueabi
 
 COPY . .
 
 ARG TARGETOS 
 ARG TARGETARCH
-RUN --mount=type=ssh \
+RUN if [ "$TARGETARCH" = "arm64" ] ; then \
```
tarakby, line 22: @sideninja or @turbolent, any chance you know how to add branching in a Dockerfile? I would like to set a different flag `CC=...` in the `go build` command depending on `TARGETARCH`, tried a few options but all were unsuccessful.

tarakby: update: no need to look at it for now, I got help and I'm investigating the next errors.

tarakby:
Builds were enabled temporarily while working on the PR. You can see the results [here](https://github.com/onflow/flow-emulator/actions/runs/7506865261/job/20439222271).

cmd/emulator/Dockerfile
```diff
@@ -20,8 +21,15 @@ ARG TARGETARCH
...
-    go build -ldflags "-extldflags -static" -o ./app ./cmd/emulator
+    if [ "$TARGETARCH" = "arm64" ] ; then \
+    GO111MODULE=on CGO_ENABLED=1 CGO_FLAGS="-O2 -D__BLST_PORTABLE__" \
+    GOOS=$TARGETOS GOARCH=$TARGETARCH CC=aarch64-linux-gnu-gcc \
+    go build -ldflags "-extldflags -static" -o ./app ./cmd/emulator ; \
+    elif [ "$TARGETARCH" = "amd64" ] ; then \
+    GO111MODULE=on CGO_ENABLED=1 CGO_FLAGS="-O2 -D__BLST_PORTABLE__" \
+    GOOS=$TARGETOS GOARCH=$TARGETARCH CC=x86_64-linux-gnu-gcc \
+    go build -ldflags "-extldflags -static" -o ./app ./cmd/emulator ; \
+    fi
```
turbolent, line 32: It looks like the only difference between the two commands for the two architectures is the CC flag?
Maybe just have `export CC=...` in the branches, and run the command shared between after the if-statement

turbolent (commented):
LGTM

cmd/emulator/Dockerfile
tarakby: I've tried to implement this without success:
- Docker does not persist shell's data from one `RUN` to the next
- For persistent data between `RUN` commands, we can use `ARG` or `ENV`, but you can't update them with shell on the Dockerfile itself 🤷🏼‍♂️ 

Let me know if you have ideas. Ideally, these if else statements should live in shell before calling `docker build`, but here our CI.yml relies on a tool that manages the `docker build` call.

turbolent: Ah, OK. I'm not a Docker expert, maybe ops can help with this

tarakby: maybe @sjonpaulbrown or @peterargue have an idea on how to set the `CC` flag within Docker so that it persists in the following commands?

peterargue: what about something like this?
```dockerfile
ARG TARGETOS 
ARG TARGETARCH
ENV GO111MODULE=on \
	CGO_ENABLED=1 \
	CGO_FLAGS="-O2 -D__BLST_PORTABLE__" \
	GOOS=$TARGETOS \
	GOARCH=$TARGETARCH
RUN --mount=type=ssh \
    --mount=type=cache,target=/go/pkg/mod \
    --mount=type=cache,target=/root/.cache/go-build \
    if [ "$TARGETARCH" = "arm64" ] ; then \
        export CC=aarch64-linux-gnu-gcc; \
    elif [ "$TARGETARCH" = "amd64" ] ; then \
        export CC=x86_64-linux-gnu-gcc; \
    fi; \
    go build -ldflags "-extldflags -static" -o ./app ./cmd/emulator ; \
```

tarakby: It works well! The `ENV` variables becomes env variables within the docker image but that's fine I guess. 
Thanks @peterargue!

tarakby: @turbolent the PR should be ready now, thanks!

bluesign (approved):

janezpodhostnik (approved):

PR #558, opened 2024-02-02
Differentiate in-memory DBs for blockchain snapshots

m-Peter:
Closes: https://github.com/onflow/flow-cli/issues/1376

## Description

For tools that may use multiple Emulator instances in the same process (e.g. `flow test`), creation & loading of snapshots is not isolated, due to the usage of in-memory databases in sqlite3 (https://www.sqlite.org/inmemorydb.html).

This can result in some conflicts, if snapshots with the same name are created in different instances, for example:

```bash
❌ Command Error: Execution failed:
error: panic: out of memory (7)
   --> Test:155:12
```

The sqlite3 error is misleading, but it basically means that the in-memory database already exists.
______

For contributor use:

### DIFF
--- a/emulator.go
+++ b/emulator.go
@@ -17,7 +17,9 @@
  */
 package emulator
 
-import "github.com/onflow/flow-emulator/emulator"
+import (
+	"github.com/onflow/flow-emulator/emulator"
+)
 
 func New(opts ...emulator.Option) (emulator.Emulator, error) {
 	return emulator.New(opts...)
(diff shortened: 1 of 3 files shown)


sideninja:
will this be able to load existing snapshots based on name? I believe there is a feature to load snapshot by name.

m-Peter:
@sideninja Yes, both loading & creation of snapshots is done by name. This is not affected, it is only the in-memory DB which is differentiated. I added a timestamp-based ID, to avoid conflicts which can happen in CI done through `flow test`.

sideninja (approved):

sideninja:
Ok to merge @m-Peter

m-Peter:
@sideninja Yes, that would be great :raised_hands: I have added tests as well :pray:

PR #564, opened 2024-02-07
[Access] SubscribeBlocks endpoint is missing

UlyanaAndrukhiv:
FLIP: https://github.com/onflow/flips/pull/229

## Description

This pull request :
- implements missing `SubscribeBlocks`, `SubscribeBlockHeaders` and `SubscribeBlockDigests` methods from Access API interface
- updates `RegisterAccessAPIServer`, `StateStreamBackend` and creating `backend.Config` according to new `flow-go` package version
- resolves dependencies (removed flow-archive dependency & storage/remote package) according to PR  [#524]
- fixed tests (according to [comment](https://github.com/onflow/flow-emulator/pull/524#issuecomment-1841497348))

______

PR #566, opened 2024-02-09
Add OnStart callback for emulator start

### LINKED ISSUE #565: Allow a callback to be trigger after emulator has started
### Issue To Be Solved

We need a way in other projects, like the CLI, to know when the emulator has started.

Introduce a variadic options arg (for non-breaking changes) that takes an options struct with a callback to be triggered once the emulator has completed.

chasefleming:
Closes #565 

______

For contributor use:

### DIFF
--- a/cmd/emulator/start/start.go
+++ b/cmd/emulator/start/start.go
@@ -88,7 +88,11 @@ type serviceKeyFunc func(
 	hashAlgo crypto.HashAlgorithm,
 ) (crypto.PrivateKey, crypto.SignatureAlgorithm, crypto.HashAlgorithm)
 
-func Cmd(getServiceKey serviceKeyFunc) *cobra.Command {
+type EmulatorOption struct {
+	OnStart func()
+}
+
+func Cmd(getServiceKey serviceKeyFunc, opts ...EmulatorOption) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "start",
 		Short: "Starts the Flow emulator server",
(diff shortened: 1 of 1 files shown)


turbolent:
Nice! What is the `release/v0` branch?

chasefleming:
@turbolent we need a way to still release non v1 features now that everything has been merged to master. I cut this branch so have the ability to do more releases like v0.60.0 if we need a change in the meantime. This means we'll have to cherry pick everything we do over here into a PR for the master branch as well. Don't foresee many changes though.

turbolent:
Sounds good! 👍 There's a discussion about having such a branch in #564, maybe good to coordinate and document it, so we don't end up with multiple different ones

chasefleming:
Thanks for the heads up @turbolent . Left a comment on there.

bluesign:
@chasefleming is this works? afaik start call is blocking (library we used there, graceland is blocking actually) 

I think emulator should start after the start call anyway, what was the issue there?

chasefleming:
@bluesign you're right. I need to rethink my approach. Moving this to a draft PR. Sorry, what do you mean by "start after start call." If `start` blocks, what I am trying to do is figure out a way to notify whatever called it that you can continue on a next step. e.g. deploy a contract to the emulator.

bluesign:
@chasefleming who is the caller here? who is starting emulator I mean. I think if you call Listen before Start, emulator should be ready anyway, os will make socket wait till it is handled.

chasefleming:
@bluesign In one use case, the CLI is the caller. I tried this, but unless I put a sleep to delay it the following code fails because it's not actually ready:

```go
emuErr := emulatorServer.Listen()
if emuErr != nil {
	// TODO: handle error
}

go func() {
	emulatorServer.Start()
}()
fmt.Println("Emulator server is ready")
```

Any ideas or see anything I'm doing wrong?

chasefleming:
This is addressed by this: https://github.com/onflow/flow-go-sdk/pull/578

Issue #526, opened 2023-12-06
Return computation used for scripts/transactinon on emulator


bjartek:
### Issue To Be Solved

When testing/desining scripts and transactions it is sometimes very hard to know how computation will scale as state grows. If there was a way to get this data from the emulator it would make life a lot easier. 

It is currently possible to get the meters for transactions but they are emitted as log entries and you have to parse the logs to get it out.

bluesign:
Yesterday on Discord we had a discussion about fixes metering, I think we lack some stuff from Cadence side to make this useful. I have few ideas, today planning to discuss with Cadence team.

PR #609, opened 2024-03-12
Computation reports for Cadence scripts & transactions

m-Peter:
Completes: 1st milestone of https://github.com/onflow/developer-grants/issues/178
Closes: https://github.com/onflow/flow-emulator/issues/526

## Description

Introduces a new endpoint under `http://localhost:8080/emulator/computationReport`, to display computation reports. For example:

computation-profile

______

For contributor use:

### DIFF
--- a/emulator/computation_report.go
+++ b/emulator/computation_report.go
@@ -26,7 +26,7 @@ import (
 )
 
 type ProcedureReport struct {
-	Path            string `json:"path"`
+	Path            string `json:"path,omitempty"`
 	ComputationUsed uint64 `json:"computation"`
 	// To get the computation from the intensities map, see:
 	// https://github.com/onflow/flow-go/blob/master/fvm/meter/computation_meter.go#L32-L39


emulator/computation_profile.go
```diff
@@ -0,0 +1,205 @@
...
+		MemoryEstimate:  txResult.MemoryEstimate,
+		Code:            code,
+		Arguments:       arguments,
+	}
+	cp.Transactions[txResult.TransactionID.String()] = txReport
+}
+
+// transformIntensities maps a numeric common.ComputationKind value
+// to a human-friendly string value.
+func transformIntensities(
```
janezpodhostnik, line 78: I wish flow-go already offered ComputationKind names.

emulator/computation_profile_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+	expectedIntensities := map[string]uint{
+		"FunctionInvocation":     2,
+		"GetAccountContractCode": 1,
+		"GetCode":                1,
+		"GetOrLoadProgram":       3,
+		"GetValue":               221,
+		"ResolveLocation":        1,
+		"Statement":              1,
+	}
```
janezpodhostnik, line 94: I would not compare all the values, as the test might become fragile.
I would maybe omit `GetValue` and perhaps `GetOrLoadProgram` as well.

emulator/computation_profile_test.go
```diff
@@ -0,0 +1,163 @@
...
+		"GenerateAccountLocalID": 1,
+		"GenerateUUID":           1,
+		"GetAccountContractCode": 1,
+		"GetCode":                1,
+		"GetOrLoadProgram":       3,
+		"GetValue":               2249,
+		"ResolveLocation":        1,
+		"SetValue":               2473,
+		"Statement":              11,
+		"TransferCompositeValue": 3,
```
janezpodhostnik, line 160: I would not check:
- EncodeValue
- GetValue
- SetValue
- AllocateStorageIndex
- GetOrLoadProgram

janezpodhostnik (approved):
Looks good. Some comments:

- It might be good to link to this function as an example on how to get the computation from the intensities https://github.com/onflow/flow-go/blob/master/fvm/meter/computation_meter.go#L32
- The naming: "profiling" hints to the profiling where you can see results per line of code executed which is a bit misleading. However if I understand correctly you intend to work towards that. If that is the case I'm ok with the name.

emulator/computation_profile.go
m-Peter: Unfortunately, I think this is not possible. The main problem is that `ComputationKind` is defined in Cadence repository and it uses the `string` cmd tool to generate the `String` representation for the values defined there. See: https://github.com/onflow/cadence/blob/master/runtime/common/compositekind.go#L27
Only the `ComputationKind` values defined in Cadence offer a human-readable name: https://github.com/onflow/cadence/blob/master/runtime/common/computationkind_string.go.

emulator/computation_profile_test.go
m-Peter: Good point :+1: 
Updated in https://github.com/onflow/flow-emulator/pull/609/commits/5dc4699492f9e2044349dedb25ef33dbf45eb23d

m-Peter:
@janezpodhostnik I have updated the naming pattern to `computation report` in https://github.com/onflow/flow-emulator/pull/609/commits/7a259eb77a41669fa66cc7e88babd5c54ba61f34, for types / variables / flags etc. I have also linked the relevant flow-go function for getting the computation from intensities.

PR #681, opened 2024-05-21
Add get full collection support

zhangchiqing:
To be able to resolve the dependency issue for https://github.com/onflow/flow-go/pull/5910

### DIFF
--- a/adapters/access.go
+++ b/adapters/access.go
@@ -172,6 +172,20 @@ func (a *AccessAdapter) GetCollectionByID(_ context.Context, id flowgo.Identifie
 	return collection, nil
 }
 
+func (a *AccessAdapter) GetFullCollectionByID(_ context.Context, id flowgo.Identifier) (*flowgo.Collection, error) {
+	collection, err := a.emulator.GetFullCollectionByID(id)
+	if err != nil {
+		return nil, convertError(err, codes.Internal)
+	}
+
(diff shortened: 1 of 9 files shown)


emulator/pendingBlock.go
```diff
@@ -126,6 +126,20 @@ func (b *pendingBlock) Collections() []*flowgo.LightCollection {
...
 
+func (b *pendingBlock) FullCollections() []*flowgo.Collection {
+	if len(b.transactionIDs) == 0 {
+		return []*flowgo.Collection{}
+	}
+
+	transactions := make([]*flowgo.TransactionBody, len(b.transactionIDs))
+	for i, txID := range b.transactionIDs {
+		transactions[i] = b.transactions[txID]
+	}
```
turbolent, line 137: Maybe just use `copy`?

turbolent (approved):

peterargue (approved):

emulator/pendingBlock.go
zhangchiqing: How to `copy`? It's reading a transaction by id from a map.

peterargue: I think this approach is needed to preserve the order

turbolent: Ah, sorry, I didn't realize the input was a map

PR #699, opened 2024-06-24
Update usage of tx validator interface

illia-malachyn:
Depends on https://github.com/onflow/flow-go/pull/6004


go.mod
```diff
@@ -236,3 +236,7 @@ require (
 	nhooyr.io/websocket v1.8.7 // indirect
 	rsc.io/tmplfunc v0.0.3 // indirect
 )
+
+replace github.com/onflow/flow-go v0.35.14-crescendo-preview.27 => github.com/The-K-R-O-K/flow-go v0.35.14-crescendo-preview.27.0.20240705100918-3f0e13f6872c
+
+replace github.com/onflow/flow-core-contracts/lib/go/templates v1.3.0 => github.com/The-K-R-O-K/flow-core-contracts/lib/go/templates v0.15.2-0.20240624091211-1812579d3242
```
turbolent, line 242: Please merge changes upstream and remove these replace statements

peterargue:
@illia-malachyn I tagged your changes from the core contracts repo `v1.3.1`. please update those here.

peterargue (approved):

illia-malachyn:
@turbolent Can we merge it in?

go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/logrusorgru/aurora v2.0.3+incompatible
 	github.com/onflow/cadence v1.0.0-preview.38
 	github.com/onflow/crypto v0.25.1
-	github.com/onflow/flow-core-contracts/lib/go/templates v1.3.0
+	github.com/onflow/flow-core-contracts/lib/go/templates v1.3.1
```
turbolent, line 19: Why are the core contracts updated? This should match what is in flow-go. If the update is needed, flow-go should be updated first

go.mod
```diff
@@ -236,3 +236,5 @@ require (
 	nhooyr.io/websocket v1.8.7 // indirect
 	rsc.io/tmplfunc v0.0.3 // indirect
 )
+
+replace github.com/onflow/flow-go v0.36.2-0.20240717214129-9ea6faeee3e7 => github.com/The-K-R-O-K/flow-go v0.35.14-crescendo-preview.27.0.20240718153829-58850e2c4673
```
turbolent, line 240: As mentioned in the other PRs, it would be great to avoid `replace` statements in `master`, as that effectively blocks all other PRs until it is removed again. It's error prone as we might accidentally tag a release with it and update downstream dependencies, where the replace statement is than _not_ in affect, leading to code with different versions

go.mod
peterargue: it's updated here:
https://github.com/onflow/flow-go/pull/6004/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R53-R54

go.mod
turbolent: Discussed this offline: Unfortunately, at the momement we do can't come up with a way to avoid replace statements for PRs from forks. So please go ahead, but remove the statement as soon as possible after the merge.

turbolent (approved):

PR #717, opened 2024-07-23
Add support for `EVM.heartbeat` resource on system chunk transaction

m-Peter:
Depends on #719
## Description

The emission of `EVM.BlockExecuted` events is now part of the system chunk transaction.
In order for tools using the Flow Emulator, to function properly, we need to handle this in Emulator's system chunk transaction as well.
Furthermore, since this system chunk transaction is rather special, we need to make its event/s available to APIs such as the event streaming.

______

For contributor use:

### DIFF
--- a/emulator/blockchain.go
+++ b/emulator/blockchain.go
@@ -1801,6 +1801,8 @@ func (b *Blockchain) systemChunkTransaction() (*flowgo.TransactionBody, error) {
 			RandomBeaconHistoryAddress: b.GetChain().ServiceAddress().Hex(),
 		},
 	)
+	// We should move this to `templates.Environment` struct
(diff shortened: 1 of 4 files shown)


sideninja:
@turbolent can you check why the migration test breaks, Ardit disabled it for the sake of referencing this update flow-emulator in EVM GW but we should resolve the issue. I tried recreating a new snapshot for the migration test which has the EVM heartbeat updated contract, but still no luck, I'm not sure which version I should use when generating the snapshot so I used v0.62.1 which I'm not sure if correct. Can you help with this

turbolent:
I improved the EVM migration in flow-go:
- https://github.com/onflow/flow-go/pull/6249
- https://github.com/onflow/flow-go/pull/6251

Once those PRs are merged, we should bump the flow-go version and then this PR should pass. Testing it locally using a temporary replace worked.

turbolent (approved):

PR #744, opened 2024-09-03
[Access] Implemented options to fail or warn when payer is not able to pay

Guitarheroua:
https://github.com/onflow/flow-go/issues/6139

Added implementation of PayerBalanceMode to control whether transaction validation should fail or warn when the payer is not able to pay.


bluesign (approved):

janezpodhostnik (commented):
good to go from my side when the replace is removed

PR #787, opened 2025-01-30
Update comments about storefront

joshuahannan:
## Description

Removes mentions of storefront from README and comments
______


bluesign:
hmm should we also remove from deploy?

PR #823, opened 2025-05-22
Scheduled callbacks

devbugging:
Closes #???

## Description

______

For contributor use:

### DIFF
--- a/emulator/callbacks.go
+++ b/emulator/callbacks.go
@@ -1,0 +1,68 @@
+package emulator
+
+import (
+	_ "embed"
+	"encoding/binary"
+	"fmt"
+	"github.com/onflow/cadence"
+	jsoncdc "github.com/onflow/cadence/encoding/json"
+	flowsdk "github.com/onflow/flow-go-sdk"
+	"github.com/onflow/flow-go/model/flow"
+)
+
+//go:embed templates/executeCallbackTransaction.cdc
+var executeCallbackScript []byte
+
+//go:embed templates/scheduleCallbackTransaction.cdc
+var scheduleCallbackScript []byte
+
+func scheduleCallbackTransaction() flow.TransactionBody {
+	return *flow.NewTransactionBody().
+		SetScript(scheduleCallbackScript)
+}
+
+func executeCallbackTransactions(scheduleEvent []flowsdk.Event) ([]flow.TransactionBody, error) {
+	var transactions []flow.TransactionBody
+
+	for _, e := range scheduleEvent {
+		limit, id, err := parseSchedulerProcessedEvent(e)
+		if err != nil {
+			return nil, err
+		}
+
+		tx := flow.NewTransactionBody().
+			SetScript(executeCallbackScript).
+			AddArgument(id).
+			SetComputeLimit(limit)
+
+		transactions = append(transactions, *tx)
+	}
+
+	return transactions, nil
+}
+
+// parseSchedulerProcessedEvent parses flow event that is emitted during scheduler
+// marking the callback as scheduled.
+// Returns:
+// - execution effort
+// - ID of the event encoded as bytes
+// - error in case the event type is not correct
+func parseSchedulerProcessedEvent(event flowsdk.Event) (uint64, []byte, error) {
+	if event.Type != "A.Scheduler.CallbackProcessed" {
+		return 0, nil, fmt.Errorf("invalid event type: %s", event.Type)
+	}
+
+	id := event.Value.SearchFieldByName("ID")
+	encodedID, err := jsoncdc.Encode(id)
+	if err != nil {
+		return 0, nil, err
+	}
+
+	effort, ok := event.Value.SearchFieldByName("computationEffort").(cadence.UInt64)
+	if !ok {
+		return 0, nil, fmt.Errorf("invalid effort value type: %v", effort)
+	}
+	computeLimit := binary.BigEndian.Uint64(effort.ToBigEndianBytes())
+
+	return computeLimit, encodedID, nil
+}
--- a/emulator/blockchain.go
+++ b/emulator/blockchain.go
@@ -1265,6 +1265,29 @@ func (b *Blockchain) executeBlock() ([]*types.TransactionResult, error) {
 		results = append(results, result)
 	}
 
+	// add transaction to schedule callbacks
+	b.pendingBlock.AddTransaction(scheduleCallbackTransaction())
+	result, err := b.executeNextTransaction(blockContext)
+	if err != nil {
+		return results, err
+	}
+	results = append(results, result)
+
+	// execute callbacks we receive from events of the schedule transaction
+	executeTxs, err := executeCallbackTransactions(result.Events)
+	if err != nil {
+		return results, err
+	}
+
+	for _, tx := range executeTxs {
+		b.pendingBlock.AddTransaction(tx)
+		result, err := b.executeNextTransaction(blockContext)
+		if err != nil {
+			return results, err
+		}
+		results = append(results, result)
+	}
+
 	return results, nil
 }
 
--- a/emulator/templates/executeCallbackTransaction.cdc
+++ b/emulator/templates/executeCallbackTransaction.cdc
@@ -1,0 +1,7 @@
+import "CallbackScheduler"
+
+transaction(id: UInt64) {
+    execute {
+        CallbackScheduler.execute(id: id)
+    }
+}
--- a/emulator/templates/scheduleCallbackTransaction.cdc
+++ b/emulator/templates/scheduleCallbackTransaction.cdc
@@ -1,0 +1,7 @@
+import "CallbackScheduler"
+
+transaction {
+    execute {
+        CallbackScheduler.process()
+    }
+}


emulator/callbacks.go
```diff
@@ -0,0 +1,126 @@
...
+	effort, ok := event.Value.SearchFieldByName(executionField).(cadence.UInt64)
+	if !ok {
+		return 0, nil, fmt.Errorf("invalid effort value type: %v", effort)
+	}
+	computeLimit := binary.BigEndian.Uint64(effort.ToBigEndianBytes())
+
+	return computeLimit, encodedID, nil
+}
+
+func replaceSchedulerAddress(script []byte, serviceAddress flow.Address) []byte {
```
chasefleming, line 118: Do we have a util for this?

emulator/callbacks.go
```diff
@@ -0,0 +1,126 @@
...
+	encodedID, err := jsoncdc.Encode(id)
+	if err != nil {
+		return 0, nil, err
+	}
+
+	effort, ok := event.Value.SearchFieldByName(executionField).(cadence.UInt64)
+	if !ok {
+		return 0, nil, fmt.Errorf("invalid effort value type: %v", effort)
+	}
+	computeLimit := binary.BigEndian.Uint64(effort.ToBigEndianBytes())
```
chasefleming, line 113: Can `efford. ToBigEndianBytes()` panic?

emulator/callbacks.go
devbugging: This will all be moved to templates as soon as we have the scheduler as part of core-contracts, so this is all temporary.

emulator/callbacks.go
devbugging: it shouldn't, the check above should make sure the value is correct

emulator/callbacks.go
```diff
@@ -0,0 +1,126 @@
+/*
```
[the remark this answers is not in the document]
devbugging: I've added tests for parsing events which is the most important part to be tested now, but other tests I would add after we make this officially relased, now it's in experimental and we will likely change things around soon

devbugging:
@chasefleming added changes.

emulator/callbacks_test.go
```diff
@@ -0,0 +1,241 @@
...
+		{
+			name: "valid event with ID=1 and effort=1000",
+			event: flowsdk.Event{
+				Type: expectedEventType,
+				Value: cadence.NewEvent([]cadence.Value{
+					cadence.NewUInt64(1),    // ID
+					cadence.NewUInt64(1000), // executionEffort
+				}).WithType(cadence.NewEventType(
+					nil,
+					"A.01.UnsafeCallbackScheduler.CallbackProcessed",
```
turbolent, line 55: Here and in the other test cases below: The type ID looks incorrect, the address part should be full length

emulator/callbacks_test.go
```diff
@@ -0,0 +1,241 @@
...
+		errorContains  string
+	}{
+		{
+			name: "valid event with ID=1 and effort=1000",
+			event: flowsdk.Event{
+				Type: expectedEventType,
+				Value: cadence.NewEvent([]cadence.Value{
+					cadence.NewUInt64(1),    // ID
+					cadence.NewUInt64(1000), // executionEffort
+				}).WithType(cadence.NewEventType(
```
turbolent, line 53: The event type is always the same for all test cases, maybe refactor it out into a variable and reuse it

server/server.go
```diff
@@ -189,6 +198,15 @@ func NewEmulatorServer(logger *zerolog.Logger, conf *Config) *EmulatorServer {
 
 	if conf.WithContracts {
 		commonContracts := emulator.NewCommonContracts(chain)
+
+		// todo refactor this once the SchedulerContract is created in core-contracts
+		if conf.ScheduledCallbacksEnabled {
```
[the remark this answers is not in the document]
turbolent: Maybe at least check that if `--scheduled-callbacks` is given, also `--contracts` must be given, otherwise I guess the feature is broken?

Alternatively maybe just make `--scheduled-callbacks` imply `--contracts`, and document it in the command usage.

emulator/callbacks.go
```diff
@@ -0,0 +1,126 @@
...
+// marking the callback as scheduled.
+// Returns:
+// - execution effort
+// - ID of the event encoded as bytes
+// - error in case the event type is not correct
+func parseSchedulerProcessedEvent(event flowsdk.Event, serviceAddress flow.Address) (uint64, []byte, error) {
+	callbackScheduledEvent := fmt.Sprintf(
+		"A.%s.UnsafeCallbackScheduler.CallbackProcessed",
+		serviceAddress.String(),
+	)
```
turbolent, line 89: Same as above: Use the functions provided by Cadence to construct a correct type ID, don't use string formatting

turbolent (commented):
Nice!

server/server.go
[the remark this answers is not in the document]
devbugging: I did it this way to be explicit during experimental phase (which will be until we have safe callback scheduler contract). But I agree that having contracts included when scheduled-callbacks is used makes sense.

emulator/callbacks_test.go
```diff
@@ -0,0 +1,246 @@
...
+ * limitations under the License.
+ */
+
+package emulator
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/onflow/cadence"
```
nialexsan, line 25:
```suggestion
	"github.com/onflow/cadence"
	"github.com/onflow/cadence/common"
```

emulator/callbacks_test.go
```diff
@@ -0,0 +1,247 @@
...
+)
+
+func TestParseSchedulerProcessedEvent(t *testing.T) {
+	serviceAddress := flow.HexToAddress("0x01")
+	const contractName = "UnsafeCallbackScheduler"
+	contractLocation := common.AddressLocation{
+		Address: common.Address(serviceAddress),
+		Name:    contractName,
+	}
+	expectedEventType := contractLocation.TypeID(nil, contractName+".CallbackProcessed")
```
nialexsan, line 41:
```suggestion
	expectedEventType := string(contractLocation.TypeID(nil, contractName+".CallbackProcessed"))
```

emulator/callbacks.go
```diff
@@ -0,0 +1,136 @@
...
+	const (
+		IDField        = "ID"
+		executionField = "executionEffort"
+	)
+
+	if event.Type != string(callbackScheduledEvent) {
+		return 0, nil, fmt.Errorf("invalid event type, got: %s, expected: %s", event.Type, callbackScheduledEvent)
+	}
+
+	id, ok := event.Value.SearchFieldByName(IDField).(cadence.UInt64)
```
turbolent, line 109: Nit: Instead of searching linearly for two fields, maybe use `FieldsMappedByName` instead and look up the fields in the result

emulator/callbacks_test.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestParseSchedulerProcessedEvent(t *testing.T) {
+	serviceAddress := flow.HexToAddress("0x01")
+	contractLocation := common.AddressLocation{
+		Address: common.Address(serviceAddress),
+		Name:    contractName,
+	}
+	eventName := fmt.Sprintf("A.%s.%s.%s", serviceAddress.String(), contractName, callbackProcessedEvent)
```
turbolent, line 40: Remove and use `expectedEventType` instead

```suggestion
```

turbolent (commented):
Nice!

chasefleming:
From another chat, it sounds like we need to upgrade the contract to the latest version.

joshuahannan:
I just pushed changes to update to the latest version of the contract and updated the tests as well. There will be future versions of the contract with significant changes, but this at least contains the basic functionality.

joshuahannan:
@devbugging Can you elaborate more on why you think it might be an issue? The current version of the contract has all the most important functionality and the same API as the old one and has a decent number of unit tests already. I'm not sure why we'd want to use one that doesn't have full functionality. Maybe it isn't clear to me what the differences are that you're talking about

devbugging:
@joshuahannan the contract tests make sure the contract works but doesn't check this integration. I'm fine if we use the new contract, but then we need to update to also use correct templates from flow-core-contract as well as the contract imported from there. So we either get in what we have and it's tested and make another PR that updates everything, or we update the emulator to use flow-core-contracts as it should and we must at least once run happy-path manually. I don't feel confident merging something noone tested. Either way it's fine with me, but I don't have now time to update the emulator to use flow-core-contracts. Maybe next week or so. If you have then that's great. Otherwise that was my concern with delays.

joshuahannan:
@jribbink Gregor told me that there were some issues when he was manually testing it so I don't know if it will work. I didn't understand the issues though. You might need to reach out to him to see what needs to be done because I don't think he gets notifications from github comments

devbugging:
@jribbink 
I tested the emulator now and after scheduled callback should be exeutor the emulator fails to produce new blocks with error:
```
❌ Command Error: failed to submit transaction: client: rpc error: code = Internal desc = pending block with ID 88bd8ce5460741226dec41f2cc4021555635e1be68ff272b8f7cd642ffcdc537 is currently being executed
```

I don't think this PR is currently in working condition. So we either revert back to my last commit which was a working experiemntal version since 2 months ago, or we update the emulator to work with current version, but I don't know when I will be able to spend time on updating emulator since my time is now focused on making flow-go work. 
That was my intention with this emulator version anyway, it wasn't to be latest version of WIP scheduled callbacks because that just adds on my plate to then maintain it. 
I can hopefully spend some time second half of the week to udpate it, but I can only do that after I finish some localnet testing, which I can't predict how long it will take.

PR #827, opened 2025-06-20
Respect config values for EVM VM bridge setup

m-Peter:
## Description

The default values for `SetupEVMEnabled` & `SetupVMBridgeEnabled` are `true`.
However, if for whatever reason, someone tries to set `false` for these two, it isn't possible to do so.
Because the option is only appended when `conf.SetupEVMEnabled` & `conf.SetupVMBridgeEnabled` are set to `true`.
______

For contributor use:


m-Peter:
@bluesign You are indeed right :100: 

I was merely trying to solve a bug in the first place, to fix the CI for https://github.com/onflow/flow-evm-gateway/pull/832 .
I haven't followed along on these VM Bridge changes, so I'll wait for @jribbink to jump in, as he may have more context on how to properly proceed with the changes.

PR #883, opened 2025-10-14
Remote Network Forking Improvements

### LINKED ISSUE #885: Align fork config with developer expectations
### Issue To Be Solved

The current flags related to forking are hidden and somewhat exotic.  While intuitive, they do not follow the norms developers expect from tools such as [foundry](https://getfoundry.sh/forge/tests/fork-testing).  Additionally, requiring the chain ID is redundant when the fork url is provided.

Rename forking flags to `--fork-host` and `--fork-start-height`.

### (Optional): Context

Productizing forking feature and incorporating into recommended DX workflows.

jribbink:
Closes #885 

## Description

Renames the flags to `--fork-host` & `--fork-height`, improves the remote store's reliability/performance for production use in CI (lru cache + rate limit management).

______

For contributor use:

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -25,6 +25,7 @@ import (
 	"net/http"
 	"os"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/onflow/cadence"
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -383,16 +411,16 @@ func configureStorage(logger *zerolog.Logger, conf *Config) (storageProvider sto
 		}
 	}
 
-	if conf.ChainID == flowgo.Testnet || conf.ChainID == flowgo.Mainnet {
+	if conf.ForkHost != "" {
```
mfbz: Should we add some kind of validation to ForkHost string like missing port, url scheme etc? it will fail regardless, but at least in this way we can notify the user with a proper error message

PR #1006, opened 2026-06-01
bump flow-core-contracts to v1.10.3

vishalchangrani:
## Summary
- Bumps `github.com/onflow/flow-core-contracts/lib/go/contracts` and `lib/go/templates` from `v1.10.2` to `v1.10.3`

## What is in v1.10.3
- Port internal fixes by @SupunS
- Fix wildcard import by @turbolent

## Test plan
- [ ] CI passes

Generated with Claude Code


turbolent (changes requested):
This should be updated in flow-go, then will be picked up by the emulator in a dependency update

